PR #2, opened 2015-12-23
Fixes "could not be converted to string" on normal use case.

andrewevansmith:
The magic getter was trying to get the content under $field->{en-US} using an object instead of the locale code. 

Had to patch this to get a simple blog post to display it's title properly.

### DIFF
--- a/src/Delivery/DynamicEntry.php
+++ b/src/Delivery/DynamicEntry.php
@@ -91,7 +91,7 @@ class DynamicEntry extends LocalizedResource implements EntryInterface
                 $locale = $this->getSpace()->getDefaultLocale();
             }
 
-            $result = $value->$locale;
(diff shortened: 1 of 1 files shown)


realityking:
Thank you for the pull request. First of all sorry for taking so long to get to this, I didn't check github over the holidays.

The issue is with fields that are not localized (which I annoyingly hadn't tested). Unfortunately your fix would have broken localized fields. I fixed the issue in a [different way](https://github.com/contentful/contentful.php/commit/538a3df395fb65c3dfc33ced9357f4c1613b6ad8#diff-883aa58f6a6c5727f30e5947d6d077ecL91) and added unit tests to ensure it doesn't happen again.

I'll cut a new release either this week or next week.

andrewevansmith:
Thanks!  :)

PR #6, opened 2016-06-23
Use TimerInterface instead of timer implementation

ueuktfd:
Here was a specific implementation of timer declared.
Timer definition was changed to interface.

### DIFF
--- a/src/Log/LoggerInterface.php
+++ b/src/Log/LoggerInterface.php
@@ -29,5 +29,5 @@ interface LoggerInterface
      *
      * @return void
      */
-    public function log($api, RequestInterface $request, StandardTimer $timer, ResponseInterface $response = null, \Exception $exception = null);
+    public function log($api, RequestInterface $request, TimerInterface $timer, ResponseInterface $response = null, \Exception $exception = null);
 }


realityking:
You're completely right, thanks for spotting this. The fix is a bit more involved then what you have submitted, if you don't want to resubmit it, I'll take care of it soon.

ueuktfd:
I'm thinking to fork your SDK and rework all log system to add support of standard PSR3 logger. 
I see you're using guzzle for http requests and this library has everything for log requests and timing. 
For each guzzle send request we can pass special option GuzzleHttp\RequestOptions::ON_STATS with closure which will log all required data. 
For example

realityking:
Fair point on `GuzzleHttp\TransferStats`, right now the SDK is sorta in-between embracing Guzzle and sticking to PSR-7. Frankly, we haven't quite made up our mind what to do in this regard. Note the SDK is still in beta, partly because we haven't made these decision. We're still trying to learn what would best serve our users.

A PSR-3 logger is also something I've considered, my plan would be to implement on top of the `LogInterface`. For some integrations we want more insight then simple string logging, so just PSR-3 (without request/response capturing) is not a direction the SDK will take.

PR #8, opened 2016-06-24
Implemented setters() and getters() for ResourceBuilder and InstanceCache

ArondeParon:
This allows for better extension of these objects, allowing developers to implement their own return types based on the Contentful API.

### DIFF
--- a/src/Delivery/Client.php
+++ b/src/Delivery/Client.php
@@ -27,17 +27,22 @@ class Client extends BaseClient
     /**
      * @var ResourceBuilder
      */
-    private $builder;
+    protected $builder;
 
     /**
      * @var InstanceCache
      */
-    private $instanceCache;
+    protected $instanceCache;
 
     /**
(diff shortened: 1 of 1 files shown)


Jalle19:
The `ResourceBuilder` is also needed to manually construct entry objects when parsing webhook payloads. Currently we have to make the object manually, which means we have to feed it a client and a cache instance etc., which is definitely not ideal.

realityking:
@Jalle19 I admittedly have not tried it but wouldn't you be able to cover webhooks with `Client::reviveJson`? It would only work for the publish and unpublished events, other things might come in a format that's not understood by the SDK.

Jalle19:
@realityking I'll check it out, hadn't seen that method before (possibly due to the slightly unobvious name)

dborsatto:
Closing this because I'm planning to tackle the extensibility issue early 2018, but in a different way.

PR #12, opened 2016-07-27
Implement fallback locales

realityking:
Implementation for the upcoming fallback locale feature.

### DIFF
--- a/src/Delivery/DynamicEntry.php
+++ b/src/Delivery/DynamicEntry.php
@@ -113,6 +113,20 @@ class DynamicEntry extends LocalizedResource implements EntryInterface
         $value = $this->fields->$fieldName;
         if (!$fieldConfig->isLocalized()) {
             $locale = $this->getSpace()->getDefaultLocale()->getCode();
+        } else {
+            $loopCounter = 0;
+            while (!isset($value->$locale)) {
+                $locale = $this->getSpace()->getLocale($locale)->getFallbackCode();
+                if ($locale === null) {
+                    // We've reach the end of the fallback chain and there's no value
+                    return null;
+                }
+                $loopCounter++;
+                // The number is arbitrary
+                if ($loopCounter > 128) {
+                    throw new \RuntimeException('Possible endless loop when trying to walk the locale fallback chain.');
+                }
+            }
         }
 
         $result = $value->$locale;
--- a/tests/Integration/FallbackLocaleTest.php
+++ b/tests/Integration/FallbackLocaleTest.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * @copyright 2015-2016 Contentful GmbH
+ * @license   MIT
+ */
+
+namespace Contentful\Tests\Integration;
+
+use Contentful\Delivery\Client;
+
+/**
+ * The fallback chain set up is the following
+ * af -> zu-ZA -> en-US -> NULL
+ * ne-NP -> NULL
+ * bs_BA -> en-US -> NULL
+ */
+class FallbackLocaleTest extends \PHPUnit_Framework_TestCase
+{
+    private function createClient()
+    {
+        $client = new Client('irrelevant', '7dh3w86is8ls');
+
+        // First we load the Space
+        $client->reviveJson('{"sys": {"type": "Space","id": "7dh3w86is8ls"},"name": "Fallback Locales","locales": [{"code": "en-US","default": true,"name": "U.S. English","fallbackCode": null},{"code": "af","default": false,"name": "Afrikaans","fallbackCode": "zu-ZA"},{"code": "zu-ZA","default": false,"name": "Zulu (South Africa)","fallbackCode": "en-US"},{"code": "bs-BA","default": false,"name": "Bosnian (Bosnia and Herzegovina)","fallbackCode": "en-US"},{"code": "ne-NP","default": false,"name": "Nepali (Nepal)","fallbackCode": null}]}');
+
+        // then the Content Type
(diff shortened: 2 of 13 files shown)


src/Delivery/DynamicEntry.php
```diff
@@ -69,9 +69,15 @@ public function getContentType()
...
+    /**
+     * @param  string $name
+     * @param  array  $arguments
+     *
+     * @return mixed
+     */
     public function __call($name, $arguments)
     {
-        if (substr($name, 0, 3) !== 'get') {
+        if (0 !== strpos($name, 'get')) {
```
mariobodemann: So _name_ can be `getter`, `target` or `together`? I think I liked the former solution better ...

src/Delivery/DynamicEntry.php
```diff
@@ -88,51 +94,73 @@ public function __call($name, $arguments)
             $getId = true;
         }
 
-        if ($fieldConfig !== null && !$fieldConfig->isDisabled()) {
-            if (!isset($this->fields->$fieldName)) {
-                if ($fieldConfig->getType() === 'Array') {
-                    return [];
-                }
+        if ($fieldConfig === null || $fieldConfig->isDisabled()) {
+            trigger_error('Call to undefined method ' . __CLASS__ . '::' . $name . '()', E_USER_ERROR);
```
mariobodemann: I think this error message is missleading. Shouldn't it be something like `disabled or non existing field found`?

src/Delivery/DynamicEntry.php
```diff
@@ -88,51 +94,73 @@ public function __call($name, $arguments)
...
-            }
+            return null;
+        }
 
-            $result = $value->$locale;
-            if ($getId && !($fieldConfig->getType() === 'Link' || ($fieldConfig->getType() === 'Array' && $fieldConfig->getItemsType() === 'Link'))) {
-                trigger_error('Call to undefined method ' . __CLASS__ . '::' . $name . '()', E_USER_ERROR);
-            }
+        if ($getId && !($fieldConfig->getType() === 'Link' || ($fieldConfig->getType() === 'Array' && $fieldConfig->getItemsType() === 'Link'))) {
+            trigger_error('Call to undefined method ' . __CLASS__ . '::' . $name . '()', E_USER_ERROR);
```
mariobodemann: Same here. It feels like a copy paste error message.

src/Delivery/DynamicEntry.php
```diff
@@ -88,51 +94,73 @@ public function __call($name, $arguments)
...
+        }
 
-            if ($getId && $fieldConfig->getType() === 'Link') {
-                return $result->getId();
+        $value = $this->fields->$fieldName;
+        if (!$fieldConfig->isLocalized()) {
+            $locale = $this->getSpace()->getDefaultLocale()->getCode();
+        } else {
+            while (!isset($value->$locale)) {
+                $locale = $this->getSpace()->getLocale($locale)->getFallbackCode();
```
mariobodemann: Is the `getSpace` part cached, or will this actually trigger a request?

src/Delivery/DynamicEntry.php
```diff
@@ -88,51 +94,73 @@ public function __call($name, $arguments)
...
+            trigger_error('Call to undefined method ' . __CLASS__ . '::' . $name . '()', E_USER_ERROR);
+        }
 
-            if ($getId && $fieldConfig->getType() === 'Link') {
-                return $result->getId();
+        $value = $this->fields->$fieldName;
+        if (!$fieldConfig->isLocalized()) {
+            $locale = $this->getSpace()->getDefaultLocale()->getCode();
+        } else {
+            while (!isset($value->$locale)) {
```
mariobodemann: Do you want to think about an endless loop protection? A counter or something, that you don't spin for ever, if the input data is _invalid_?

tests/Integration/FallbackLocaleTest.php
```diff
@@ -0,0 +1,90 @@
...
+        $client->reviveJson('{"sys": {"type": "Space","id": "7dh3w86is8ls"},"name": "Fallback Locales","locales": [{"code": "en-US","default": true,"name": "U.S. English","fallbackCode": null},{"code": "af","default": false,"name": "Afrikaans","fallbackCode": "zu-ZA"},{"code": "zu-ZA","default": false,"name": "Zulu (South Africa)","fallbackCode": "en-US"},{"code": "bs-BA","default": false,"name": "Bosnian (Bosnia and Herzegovina)","fallbackCode": "en-US"},{"code": "ne-NP","default": false,"name": "Nepali (Nepal)","fallbackCode": null}]}');
+
+        // then the Content Type
+        $client->reviveJson('{"sys": {"space": {"sys": {"type": "Link","linkType": "Space","id": "7dh3w86is8ls"}},"id": "sampleType","type": "ContentType","createdAt": "2016-06-21T08:53:55.767Z","updatedAt": "2016-06-21T13:16:34.507Z","revision": 5},"displayField": "title","name": "Sample Type","description": "","fields": [{"id": "title","name": "title","type": "Symbol","localized": true,"required": true,"disabled": false,"omitted": false}]}');
+
+        return $client;
+    }
+
+    /**
+     * No ne-NP
```
mariobodemann: I'm now not anymore sure, if I like the naming of the locales here. Do you mind writing what no ne-NP means?

src/Delivery/DynamicEntry.php
realityking: It's cached by the client.

src/Delivery/DynamicEntry.php
realityking: It's not perfect, but the old one did that too.

src/Delivery/DynamicEntry.php
realityking: `__call()` is the magic method invoked when the method called could not be found. The error message is emulating the one from the PHP engine when a method is missing, as we swallow that error by declaring `__call()`.

PR #19, opened 2016-09-13
fixed an issue if a DynamicEntry has a oneToMany link but has hanging…

graymic:
We added this to fix an issue we are facing when an entry has a child link that is deleted:

Steps to replicate: 

1. Create entry with oneToMany reference.
2. Create two entries and assign to that reference.
3. Delete one of two entries.
4. Try and retrieve that field $entry->getOneToMany(), it will fail because there is a hanging reference of a previously deleted entry, making it impossible to retrieve the one does exist.

### DIFF
--- a/src/Delivery/Client.php
+++ b/src/Delivery/Client.php
@@ -10,6 +10,7 @@ use Contentful\Client as BaseClient;
 use Contentful\Delivery\Synchronization\Manager;
 use Contentful\Query as BaseQuery;
 use Contentful\Log\LoggerInterface;
+use Contentful\ResourceNotFoundException;
 
 /**
  * A Client is used to communicate the Contentful Delivery API.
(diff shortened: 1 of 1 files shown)


realityking:
Hi Michael,

you bring up an interesting problem that we haven't considered before. I have, however, a few issues with the concrete implementation:

- It's affecting all links, not just those in arrays.
- It returns an empty array as the placeholder while everything else are objects
- There's no indication to the developer that there's anything wrong with the reference

The first two points can easily be addressed by moving this logic to DynamicEntry::_call (specially this line https://github.com/contentful/contentful.php/blob/79167e96e3922f5aa288a3258b674f8905726c41/src/Delivery/DynamicEntry.php#L124). For the the last point I'm not quite sure. Do you have any ideas?

Cheers,
Rouven

graymic:
Hi @realityking, I'm happy to move it to that line and agree that's a better location for it from what I can see. 

Firstly, I'll check locally if this fixes the issue we're facing and has no impact elsewhere. 

RE the last point, I'd have to go away and do some research on it because nothing obvious springs to mind other than another wrapper class unique to that scenario. I am worried that wouldn't really be solving the issue per se.

realityking:
Hey @graymic, sorry it took me so long to get back to this. I implemented it as discussed in #30. Could you please confirm that this one work for you?

graymic:
Hey @realityking  awesome!

I can confirm this now works within our environment and I'm able to catch the exceptions accordingly and strip out fields where required.

Thanks, 

Graymic

Issue #7, opened 2016-06-24
Don't use private properties on the Delivery Client


ArondeParon:
I'm currently trying to integrate the Contentful SDK into a Laravel application and want to extend the Delivery Client and Resource Builder objects in order to make them return custom objects instead of the ResourceArray and DynamicEntry objects.

This requires extending the DeliveryClient to create a custom Builder, which is not possible because of the fact that the `$builder`, `$instanceCache` and `$preview` properties are private.

Please make these properties `protected` so that we can build on them :)

An alternative, in my opinion even better, solution would be implement `createResourceBuilder()` and `createInstanceCache()` methods that can be extended.

ArondeParon:
Added a pull request to demonstrate this.

https://github.com/contentful/contentful.php/pull/8

realityking:
Hi Aron,

the extension story for the SDK is admittedly somewhat unrefined. Long term the plan is to allow you to override the `ResourceBuilder`, whether that's through a factory method or dependency injection is not quite clear. The reason why you can't do that today is that the interface of the `ResourceBuilder` is far from stable.

The other part of this is that we want to allow generating custom classes to use with the `ResourceBuilder` [see the ClassGenerator](https://github.com/contentful/contentful.php/blob/63b0b09cca0cd43ce002640d06c9faa0335e9470/src/Delivery/Tool/ClassGenerator.php). That's not yet enabled since the generated classes are still buggy and I haven't gotten around to fixing them. Customising them would also rely on a bit more stable interface for `ResourceBuilder`.

I'd be curios why you're trying to override `ResourceArray`. Maybe you could share a bit more about what you're trying to achieve?

Cheers,
Rouven

ArondeParon:
Hi Rouven,

Thanks for the speedy response. I'll clarify my use-case a bit. I'm using Contentful as the backend for a Laravel-driven website and want to model my data using simple models. I made a `ContentfulModel` that talks to a custom query builder. Reason for doing this is that I can now define and retrieve my content like this:

```
class BlogItem extends ContentfulModel
{
    protected $source = 'blogItem';
}
```

Which can then be used to retrieve content using `BlogItem::all()` or queries like `BlogItem::where('fields.slug', 'foo')->get()`. This make my code a lot more expressive.

The problem, however, is that the `get()` method eventually proxies to `Contentful\Delivery\Client::get` which will then return a `ResourceArray`. In *my* ideal situation, it would return a `BlogItem` model.

[...]

PR #21, opened 2016-09-22
added getFields to retrieve fields

etwillbefine:
Added the `getFields` method on DynamicEntry to retrieve fields. This gives us the opportunity to check whether properties exist or not. The problem ist, that currently I don't see any chance to keep track on non existent properties (no "magic method" should be called in that case). 

Would be great to merge that.
Thanks.

Otherwise we could add a method to check for existing properties (eg: `propertyExist($name)`)
/cc @realityking

### DIFF
--- a/src/Delivery/DynamicEntry.php
+++ b/src/Delivery/DynamicEntry.php
@@ -69,6 +69,11 @@ class DynamicEntry extends LocalizedResource implements EntryInterface
         return $this->sys->getContentType();
     }
 
+    public function getFields()
+    {
+        return $this->fields;
+    }
+
     public function __call($name, $arguments)
     {
         if (substr($name, 0, 3) !== 'get') {


etwillbefine:
Any chance to get a reply on this?

etwillbefine:
@realityking ?

Haehnchen:
+1

lukewduncan:
+1

realityking:
@etwillbefine @Haehnchen @lukewduncan I got an alternate proposal in in pull request #49. Could you check if that'd address your requirements? Thank you.

etwillbefine:
Hm imo this does not solve the issue. One of the advantages of using this sdk is the provided mapping/linking (basically the [ResourceBuilder](https://github.com/contentful/contentful.php/blob/master/src/Delivery/ResourceBuilder.php)). 
Why the user should be forced to retrieve data in raw format just to check whether a field exists or not?

realityking:
Thanks for the feedback. I think I operated under an incorrect assumption. Maybe you can clarify a bit why you're checking for properties?

Do you know the _field_ exists and you want to check if it has content? Or do you want to check if the field/property is even part of the content model?

etwillbefine:
This PR is about to check whether the field/property is part of the content model. This issue is quite old and I cannot remember the exact use case anymore. But imagine you want to extend the DynamicEntry and a content-model may contains an additional field/id/another model/whatever and you don't want to force the author to add this field to the content-MODEL (as you only want to do some stuff if its present). (these use cases assumes that you don't control the contentful cms side completely or you want to treat some objects on an abstract view)

Example:
```php
class TrackableEntry extends DynamicEntry {
    // only one or even none of these property are required in the content-model

    public function hasTrackingInformation() {
       return (woha how can I do this?)
    }
    public function hasTrackingEventName() {
       return (woha how can I do this?)
   }
}

public class EventListener {
     public function onEvent(DynamicEntry $e) {
         log.info($e->getId(), $e->getAuthorId()); 
        // uuh, triggers php error if entry doesn't have an author id attached
        // ok lets try to log it in a different way when the author-id is attached
        // woha how can I do this? 
    }
}

public class Mapper {
     public function map(DynamicEntry $e) {
         $entry = new MyCustomObject();
         $entry->id = $e->getId(); // great!
         $entry->name = $e->getName(); // fine as the name is required

         $entry->images = $e->getImages(); // oups, it may only contains `image` or `images`
         // lets check if images or image was provided, woha how can I do this?
    }
}

....
```

So as a conclusion to try to clarify the use case behind that: If the service grows and you want to unify some code for all content-types or if you want to treat a "wrong/misconfigured content-model" in a proper way you must be somehow able to look behind the scenes. 

If you don't want to make all fields accessible using this `getFields` method, as I said another way could be to add a method to check whether a property exists or not. The DynamicEntry is not very easy to extend, but imo it would be better to stay open for extension. From the point of view it is very similar to this issue https://github.com/contentful/contentful.php/issues/7

I know that `sys` is protected in the DynamicEntry, but that seems similar to retrieving data in raw format. I'd not appreciate it.

*As I did nothing related to contentful the last 3 months maybe I missed something?!*

PR #23, opened 2016-09-28
Add a editorconfig file.

lucasmichot:
- Add a editorconfig file - see http://editorconfig.org/
- Also reformat composer.json to comply editorconfig


lucasmichot:
Ping @realityking

PR #25, opened 2016-09-28
Ensure PSR-2 coding style is respected.

lucasmichot:
Ensure PSR-2 coding style is respected.

Fixing the coding-style can be achived by running: `./vendor/bin/php-cs-fixer fix`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ vendor/
 build/
 .idea/
 .DS_Store
+.php_cs.cache
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,13 @@
+<?php
+
+use Symfony\CS\Config\Config;
+use Symfony\CS\Finder\DefaultFinder;
+use Symfony\CS\FixerInterface;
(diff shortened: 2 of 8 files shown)


realityking:
Hi Lucas!

Love the fixes but I'd like to avoid adding more dependencies to `require-dev` (they tend to conflict after a while). Could you make use of the `.phar` version instead. Thanks :)

lucasmichot:
Sure, just removed the dependency in https://github.com/contentful/contentful.php/pull/25/commits/b16467baf6be401dfc2155e021afb8cf23c2e0ca

realityking:
Gee you're quick. You'll also have to remove `.php_cs` and revert `.gitignore` since those changes are now redundant.

Thanks again :)

lucasmichot:
Well in fact this is the fixing definition file that can be used using the phar version of PHP-CS
This allows to avoid typing the fixing arguments in the CLI

realityking:
I didn't know that. In this case 👍

PR #27, opened 2016-10-16
[client] : Allow httpClient to be overridden

solomonjames:
In order to enable the user to add custom middleware to the GuzzleHttp\Client object, the private class variable `httpClient` needs to be accessible for overriding.

By storing the constructor param, `$token`, in the class variable `self::$token` we can easily ensure that the `BearerToken()` middleware is pushed to the handler stack when a custom httpClient is provided.

For my use-case, I wanted to add custom GuzzleHttp middleware to cache the API results to cut-down on API calls and speed up the site.  A somewhat better option might be to do some sort of full-page caching with a reverse-proxy but that gets complicated when you have users logged into your application.

@realityking I am pinging you as you seem to be the main/only author of the Contentful\Client class.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -66,6 +66,13 @@ abstract class Client
         ]);
     }
 
+    /**
+     * Override the default httpClient
+     *
+     * @param GuzzleClient $httpClient
+     *
+     * @return self
+     */
     public function setHttpClient(GuzzleClient $httpClient)
     {
         // Ensure that the BearerToken middleware is added to the custom client
--- a/src/Client.php
+++ b/src/Client.php
@@ -27,6 +27,11 @@ abstract class Client
     /**
      * @var string
      */
+    private $token;
+
+    /**
+     * @var string
+     */
     private $baseUri;
 
     /**
@@ -51,6 +56,7 @@ abstract class Client
     {
         $stack = HandlerStack::create();
         $stack->push(new BearerToken($token));
+        $this->token = $token;
         $this->logger = $logger ?: new NullLogger();
 
         $this->api = $api;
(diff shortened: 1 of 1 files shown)


realityking:
Hi James,

I agree, this is something we need to open up. I am, however, not a fan of allowing people to change the Guzzle client after creating the object.

There are a couple of breaking changes coming up for the next release (0.7.0) we might have an opportunity to do something more daring:

- Have the constructor accept a Guzzle client.
- Have a static method that creates an instance using the default config (essentially what we're doing right now in the constructor)

Those wanting a quick setup use the static method and people with more custom needs have to endure managing a bit more of the internals themselves.

What do you think?

solomonjames:
@realityking As long as a custom Guzzle client can be used that works for me.  And as long as the constructor still ensures the BearerToken is present in the handlers, that works for me.

I personally have no issues with the setting of the client post-instantiation so that is one reason I went with this method.  It also seemed to be the easiest path, with the least code changes.

I am currently building out a site, and will be using my fork until we have something that enables the use of a custom client.

Thank you.

realityking:
> And as long as the constructor still ensures the BearerToken is present in the handlers, that works for me.

Interesting consideration, I would have skipped that part. Let me think about it a little bit more.

realityking:
After thinking about it for a while, I'll go with the approach outlined by @Haehnchen in #32 combined with what I described in an earlier comment. The bits and pieces necessary for this will land in the next few days.

solomonjames:
@realityking Just checked out #32 and that looks great and seems like a smart approach to the client independence.  What are your thoughts on me submitting a new PR that simply adds a `setHttpClient` method so it can be injected instead of my having to extend the `Client` class?

realityking:
@solomonjames I'm not a big fan of setter injection. Would the approach in #36 work for you?

Haehnchen:
@realityking go for #36  :thumbsup:

solomonjames:
@realityking @Haehnchen

I think something that is totally optional shouldn't need to clutter the constructor and works great with setter injection. Reference: http://symfony.com/doc/current/service_container/injection_types.html#setter-injection

Not sure I agree with the approach, but its better than nothing.  Thank you.

realityking:
Closing this since I finally merged #36. Thank you for the inspiration :)

PR #32, opened 2016-10-27
drop HandlerStack and BearerToken dependency to make Guzzle client more independent

Haehnchen:
Instead of provide a guzzle instance directly in Client constructor. i want to inject my own Guzzle client with attached logging eg for Symfony profiling. currently there is now easy way. this pull request drops HandlerStack. Next step will now to move client into constructor itself.

Currently i am use a hacky workaround:

``` php
    /**
     * @var string
     */
    private $token;

    public function __construct(ClientInterface $client, $token, $spaceId, $preview = false, LoggerInterface $logger = null)
    {
        parent::__construct($token, $spaceId, $preview, $logger);
        $this->token = $token;

        $binding = \Closure::bind(function() use ($client) {
            $this->httpClient = $client;
        }, $this, \Contentful\Client::class);

        $binding();
    }

    protected function request($method, $path, array $options = [])
    {
        return parent::request($method, $path, array_merge_recursive($options, [
            'headers' => ['Authorization' => 'Bearer ' . $this->token],
        ]));
    }
```


realityking:
Sorry for the long delay, I thought about this for a while since it got brought up in #32. In the end I agree, your approach is the most sensible.

PR #41, opened 2017-01-02
Add support for select api

Haehnchen:
Add support for "select" field in Query object. see: https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters/select-operator/query-entries


realityking:
This is a great start but before I'd be comfortable merging this I'll need write some tests that show the interaction of partial content with the RessourceBuilder class.

At the very least we'll probably need a list of meta properties you can never not select because they're needed for the RessourceBuilder to work. This might just be `sys.id` but I haven't looked any deeper.

realityking:
Closing in favour of #60. Thank you for the contribution :)

PR #42, opened 2017-01-06
Fixes the Assets' class handling of missing locales by implementing fallback locales.

### LINKED ISSUE #38: Asset get() methods can throw Uncaught Exceptions when the content model is non-localized
`Asset` methods such as `getFile()` and `getTitle()` can throw uncaught exceptions when the `Asset` in question has `localized` property set to `false`.

For example, say i have the regular `DynamicEntry` - `Car`, with a linked field `Image`, which is an `Asset`.

Let's say the `Image` looks like this:

```json
  {
    "name": "Image",
    "id": "image",
    "type": "Link",
    "linkType": "Asset",
    "validations": [],
    "localized": false,
    "required": false,
    "disabled": false,
    "omitted": false
  }
```

Now, if you attempt to `getTitle('fr')` on this `Asset` for example, the SDK will throw:

`Undefined property: stdClass::$fr`

I'd suggest the expected behaviour in this scenario should be fallback to the default value.

realityking:
Fixes #38

### DIFF
--- a/tests/Unit/Delivery/AssetTest.php
+++ b/tests/Unit/Delivery/AssetTest.php
@@ -36,14 +36,23 @@ class AssetTest extends \PHPUnit_Framework_TestCase
             ->getMock();
 
         $defaultLocale = new Locale('en-US', 'English (United States)', null, true);
+        $klingonLocale = new Locale('tlh', 'Klingon', 'en-US');
+        $germanLocale = new Locale('de-DE', 'German', 'en-US');
 
         $space->method('getId')
             ->willReturn('cfexampleapi');
         $space->method('getLocales')
             ->willReturn([
                 $defaultLocale,
-                new Locale('tlh', 'Klingon', 'en-US')
+                $klingonLocale,
+                $germanLocale
             ]);
+        $space->method('getLocale')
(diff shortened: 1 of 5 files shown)


tests/Unit/Delivery/AssetTest.php
```diff
@@ -85,6 +91,42 @@ public function testGetter()
         $this->assertEquals(new \DateTimeImmutable('2013-09-02T14:56:34.240Z'), $asset->getUpdatedAt());
     }
 
+    public function testGetTitleWithLocale()
+    {
+        $asset = $this->asset;
+
+        $this->assertEquals('Nyan Cat', $asset->getTitle());
+        $this->assertEquals('Nyan Cat', $asset->getTitle('en-US'));
+        $this->assertEquals('Nyan Cat', $asset->getTitle('tlh'));
```
loudmouth: should we add a test case that has a different expected value other than `'Nyan Cat'` i.e. a translated title?

realityking: Good point, let me change the tests.

realityking:
@loudmouth could you have another look?

PR #46, opened 2017-01-23
Removes the caching of Entry and Asset instances.

realityking:
The instance cache represents a form of global state. Since Entry and Asset allow
switching between locales this state has caused some confusion.

It's also incompatible with future changes that are planned, including support for the
`select` and locale `operators`.

### DIFF
--- a/src/Delivery/DynamicEntry.php
+++ b/src/Delivery/DynamicEntry.php
@@ -16,6 +16,11 @@ class DynamicEntry extends LocalizedResource implements EntryInterface
     private $fields;
 
     /**
+     * @var array
+     */
+    private $resolvedLinks = [];
+
+    /**
      * @var SystemProperties
      */
     protected $sys;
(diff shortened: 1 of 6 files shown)


src/Delivery/DynamicEntry.php
```diff
@@ -130,7 +136,15 @@ public function __call($name, $arguments)
         }
 
         if ($result instanceof Link) {
-            return $client->resolveLink($result);
+            $cacheId = $result->getLinkType() . '-' . $result->getId();
+            if (isset($this->resolvedLinks[$cacheId])) {
+                return $this->resolvedLinks[$cacheId];
+            }
+
+            $resolvedObj = $client->resolveLink($result);
```
dlitvakb: I wouldn't issue an extra request for this, and rather leave the link as is.

This is specially the case for deleted referenced entries/assets.

PR #48, opened 2017-01-25
Adds the possibility to query for a single locale.

realityking:
Also make it possible to specify a default locale when creating a client.

This is a major breaking change but adresses an often-requested feature. The new default
behavior is is to request content the space's default locale. To retain the old behavior
set the default locale to `'*''` when created the client. This could look like this:
`new Client($token, $spaceID, false, null, null, '*')`


src/Delivery/ResourceBuilder.php
```diff
@@ -169,14 +169,17 @@ private function buildArray(array $data, array $rawDataList = null)
...
+        $files = array_map([$this, 'buildFile'], $this->normalizeFieldData($fields['file'], $locale));
 
         $asset = new Asset(
-            isset($fields['title']) ? $fields['title'] : null,
-            isset($fields['description']) ? $fields['description'] : null,
+            isset($fields['title']) ? $this->normalizeFieldData($fields['title'], $locale) : null,
+            isset($fields['description']) ? $this->normalizeFieldData($fields['description'], $locale) : null,
             $files,
-            $this->buildSystemProperties($data['sys'])
+            $sys
```
loudmouth: So if I'm understanding this correctly, `buildAsset` returns an `Asset` object that only reflects the data of one locale (e.g. it has been localized)?

tests/E2E/AssetBasicTest.php
```diff
@@ -23,20 +24,43 @@ public function setUp()
...
+    }
+
+    /**
+     * @vcr e2e_asset_get_one_locale_default.json
+     */
+    public function testGetOneSingleLocale()
+    {
         $asset = $this->client->getAsset('nyancat');
 
         $this->assertInstanceOf(Asset::class, $asset);
```
loudmouth: Should we test a different single locale to make sure it's done properly? If the test is only testing default, it's hard to know if behavior is correct.

tests/E2E/EntryBasicTest.php
```diff
@@ -24,20 +25,43 @@ public function setUp()
...
+
+        $this->assertInstanceOf(ResourceArray::class, $assets);
+    }
+
+    /**
+     * @vcr e2e_entry_get_all_locale_default.json
+     */
+    public function testGetAllDefaultLocale()
+    {
         $assets = $this->client->getEntries();
```
loudmouth: I feel like you can assert some more stuff here. Just testing that it's an array doesn't test that the data is what you expect it to be. Maybe assert that the first element in the array is a certain thing with a certain localized title etc.

tests/E2E/EntryLocaleTest.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public function testGetTlh()
+    {
+        $query = (new Query)
+            ->setContentType('cat')
+            ->setLocale('tlh');
+        $entries = $this->client->getEntries($query);
+
+        $this->assertInstanceOf(ResourceArray::class, $entries);
+        $this->assertEquals('Nyan vIghro\'', $entries[0]->getName());
```
loudmouth: 👍

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 * Added support for the `webp` format in the Images API.
 * Introduced `RateLimitExceededException` for more specific error handling. **[BREAKING]**
 * Allow injecting a custom Guzzle instance into `Client`.
+* Allow fetching content in a single locale by adding the locale code to the query. **[BREAKING]**
+  **MIGRATION:** To retain the old behavior set the default locale to `'*''` when created the client. This could look
```
realityking: Both fixed.

src/Delivery/Client.php
```diff
@@ -41,17 +41,25 @@ class Client extends BaseClient
...
      * @param LoggerInterface       $logger
      * @param GuzzleClientInterface $guzzle
+     * @param string|null           $defaultLocale The default is to fetch the Space's default locale. Set to a locale
+     *                                             string, e.g. "en-US" to fetch content in that locale. Set it to "*"
+     *                                             to fetch content in all locales.
      *
      * @api
      */
-    public function __construct($token, $spaceId, $preview = false, LoggerInterface $logger = null, GuzzleClientInterface $guzzle = null)
+    public function __construct($token, $spaceId, $preview = false, LoggerInterface $logger = null, GuzzleClientInterface $guzzle = null, $defaultLocale = null)
```
realityking: `null` means it's going to use the Space's default locale. (As configured in Contentful)

src/Delivery/Client.php
```diff
@@ -76,16 +84,18 @@ protected function getUserAgentAppName()
...
-     * @param  string $id
+     * @param  string      $id
+     * @param  string|null $locale
      *
      * @return Asset
      *
      * @api
      */
-    public function getAsset($id)
+    public function getAsset($id, $locale = null)
```
realityking: PHP can't do that :(

tests/E2E/EntryBasicTest.php
realityking: The main point in these tests is that it doesn't fall over and throws an exception when constructing the classes. Though more assertions wouldn't hurt ;)

src/Delivery/Client.php
```diff
@@ -139,16 +151,18 @@ public function getContentTypes(BaseQuery $query = null)
...
+     * @param  string|null $locale
      *
      * @return EntryInterface
      *
      * @api
      */
-    public function getEntry($id)
+    public function getEntry($id, $locale = null)
     {
+        $locale = $locale === null ? $this->defaultLocale : $locale;
```
realityking: Only twice, I think that's fine ;)

src/Delivery/Client.php
```diff
@@ -163,7 +177,9 @@ public function getEntries(BaseQuery $query = null)
     {
         $query = $query !== null ? $query : new BaseQuery;
         $queryData = $query->getQueryData();
-        $queryData['locale'] = '*';
+        if (!isset($queryData['locale'])) {
+            $queryData['locale'] = $this->defaultLocale;
+        }
```
realityking: Also only twice

tests/E2E/AssetBasicTest.php
realityking: The main difference is that the data structure from the API is different for single locale vs. multi locale - those need to be tested here. The actual access in the `Asset` class is the same, as it get normalised before the class is constructed.

src/Delivery/ResourceBuilder.php
realityking: Both work. See this part: https://github.com/contentful/contentful.php/pull/48/files/66bf3c2c23eaf022ccd2ef12c9e07d221ad7713f#diff-52781e0c24eff4092e33103bf162cbd4R179

src/Delivery/Client.php
loudmouth: 👍

src/Delivery/ResourceBuilder.php
loudmouth: :+1: to empty link :-P:

PR #49, opened 2017-01-27
[WIP] Adds the ability to fetch raw ressources.

realityking:
These responses are completely unprocessed and correspond directly to the API structure.


Jalle19:
IMO what is actually needed is a way to get the raw decoded JSON out of e.g. `DynamicEntry` objects. We've hit an issue where it seems the JSON structure when "pulling" data from the API is not identical to what you get when you receive a webhook request, and the SDK currently can't handle the webhook payload correctly.

realityking:
Unfortunately by the time you have a `DynamicEntry` the raw data is long gone, the objects store them quite differently as we have to/want to normalize some things.

But I'm curios, how would have access to the raw data have helped you with #103?

PR #56, opened 2017-03-08
use psr-16 (simple cache)

daniel-rose:


dborsatto:
Hey @daniel-rose,

thanks for contributing! Though I overall like the idea, there is something that concerns me: [as per specs](http://www.php-fig.org/psr/psr-16/), an object to be able to be fully PSR-16 compatible should support lossless serialization in the form `$o == unserialize(serialize($o))`, and currently I'm not sure 100% this would be the case.

Moreover, there's a lot of boilerplate code in every `Client::getX` method, which is not ideal, and could probably be solved by a refactoring. Finally, if we were to adopt PSR-16, we could actually choose to use a package like [`cache/array-adapter`](https://packagist.org/packages/cache/array-adapter) to provide a default implementation, rather than having to maintain one (which is currently minimal because of the restricted used case).

For all these reasons (plus the fact that the branch is no longer sync'ed with master and there would be a lot of conflicts), I think I'll close this PR, but I will add further investigating PSR-16 compatibility to my backlog 🙂 👍

PR #64, opened 2017-03-21
Adds support for sync in the Preview API.

realityking:


tests/E2E/SyncTest.php
```diff
@@ -12,31 +12,36 @@
...
+     */
+    public function testPreviewSync()
+    {
+        $manager = (new Client('e5e8d4c5c122cf28fc1af3ff77d28bef78a3952957f15067bbc29f2f0dde0b50', 'cfexampleapi', true))
+            ->getSynchronizationManager();
+
+        $result = $manager->startSync();
+
+        $this->assertInstanceOf(Result::class, $result);
+        $this->assertTrue($result->isDone());
```
mariobodemann: Do you want to check here if the result is actually containig values you would expect?

realityking: I think this is ok for now. I'll add some more exhaustive tests at some point in the future.

PR #67, opened 2017-03-23
[WIP] Cache request

pontifex:
It is just a Prototype so far.


pontifex:
We finally decided to use master branch instead last release tag. Code in master allows us to configure Guzzle Client. Cache is done on request using the client and not on contents with the prototype from this PR.

realityking:
Hi @pontifex,

sorry for the lack of a response. Client-side caching is something we haven't really figured out yet, implementing your own strategy is really your best bet right now.

Cheers,
Rouven

PR #69, opened 2017-03-28
Also tests for PHP 7.1.

lucasmichot:
Also tests for PHP 7.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.5
   - 5.6
   - 7.0
+  - 7.1
   - hhvm
 
 matrix:
@@ -18,16 +19,15 @@ branches:
 
 before_install:
  - wget https://github.com/satooshi/php-coveralls/releases/download/v0.7.0/coveralls.phar
- - wget -O phpunit.phar https://phar.phpunit.de/phpunit-4.8.27.phar
 
 install:
-  - composer self-update
-  - composer install
+  - travis_retry composer self-update
+  - travis_retry composer install
 
 script:
  - vendor/bin/parallel-lint src/
  - vendor/bin/parallel-lint tests/
- - php phpunit.phar -v
+ - vendor/bin/phpunit -v
 
 after_success:
  - travis_retry php coveralls.phar -v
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -13,7 +13,8 @@
     "require-dev": {
         "php-vcr/phpunit-testlistener-vcr": "^1.1",
         "jakub-onderka/php-parallel-lint": "^0.9.2",
-        "jakub-onderka/php-console-highlighter": "^0.3.2"
+        "jakub-onderka/php-console-highlighter": "^0.3.2",
+        "phpunit/phpunit": "~4.8 || ~5.0"
```
realityking: Not a fan of this, I've had conflicts with dependencies I've considered integrating with the SDK before. Whatever can safely stay outside of composer should. (I realise that's a bit of a pain in a number of ways but pain > dependency conflicts)

Instead I'd modify this line to conditionally load a different version of phpunit. https://github.com/contentful/contentful.php/pull/69/files#diff-354f30a63fb0907d4ad57269548329e3L21

.travis.yml
```diff
@@ -18,16 +19,15 @@ branches:
 
 before_install:
  - wget https://github.com/satooshi/php-coveralls/releases/download/v0.7.0/coveralls.phar
- - wget -O phpunit.phar https://phar.phpunit.de/phpunit-4.8.27.phar
 
 install:
-  - composer self-update
-  - composer install
+  - travis_retry composer self-update
+  - travis_retry composer install
```
realityking: Interesting, did you have problems with composer failing?

realityking:
I'll see what it takes to drop PHP 5.5, not sure if that's gonna work out though.

.travis.yml
lucasmichot: It's a common practice to avoid timeouts: https://docs.travis-ci.com/user/common-build-problems/

composer.json
lucasmichot: The fact is that this dependency has actually been properly resolved by composer. see: https://travis-ci.org/contentful/contentful.php/builds/215916416

Also this allows any people to test locally PR they submit to this repo, without forcing them to install globally a specific phpunit version

lucasmichot: No same issue with phar apply here

PR #70, opened 2017-03-28
Let user run PHP-CS-Fixer 2 and fix CS

lucasmichot:
Let user run PHP-CS-Fixer 2 and fix CS

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
     "require-dev": {
         "php-vcr/phpunit-testlistener-vcr": "^1.1",
         "jakub-onderka/php-parallel-lint": "^0.9.2",
-        "jakub-onderka/php-console-highlighter": "^0.3.2"
+        "jakub-onderka/php-console-highlighter": "^0.3.2",
+        "friendsofphp/php-cs-fixer": "~2"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -13,7 +13,8 @@
     "require-dev": {
         "php-vcr/phpunit-testlistener-vcr": "^1.1",
         "jakub-onderka/php-parallel-lint": "^0.9.2",
-        "jakub-onderka/php-console-highlighter": "^0.3.2"
+        "jakub-onderka/php-console-highlighter": "^0.3.2",
+        "friendsofphp/php-cs-fixer": "~2"
```
realityking: Same as with phpunit, let's not add dependencies just for convenience's sake.

lucasmichot: Sure, nevertheless this allows people to provide consistent CS on PR they submit.
Also projects require this library use the `--no-dev` during the build process, so this dependencies will be excluded.

In this case in might be complicate for a developer to install the right version of php-cs-fixer (~1 and ~2) are not compatibles, and not force them to install a specific version globally

realityking: I agree that the 1 vs 2 will be an issue here but the rule of "anything that can be a phar should be" is not one I'm willing to compromise on. Many of the same dependencies used by these tools are also used by code generation libraries and one of those will probably end up in this SDK at some point. With PHP/composer only supporting a single version for each library I'd rather be extra careful and avoid version clashes.

lucasmichot: Sure

PR #79, opened 2017-03-30
Adds support for caching the space metadata and content types.

realityking:
The cache can only be warmed up through the CLI, it can not be iterativly build up by using the API. This is done to ensure that all content types are from one point in time.

### DIFF
--- a/src/Delivery/Cache/FilesystemCache.php
+++ b/src/Delivery/Cache/FilesystemCache.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * @copyright 2015-2017 Contentful GmbH
+ * @license   MIT
+ */
+
+namespace Contentful\Delivery\Cache;
+
+use Symfony\Component\Filesystem\Filesystem;
+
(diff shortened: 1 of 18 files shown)


src/Delivery/Cache/FilesystemCache.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * FilesystemCache constructor.
+     *
+     * @param  string $cacheDir
+     * @param  string $spaceId
+     */
+    public function __construct($cacheDir, $spaceId)
+    {
+        $this->fs = new Filesystem;
+        $this->cacheDir = $cacheDir . '/' . $spaceId;
```
Khaledgarbaya: excuse my PHP, isn't there a function to do a join instead of specifying `'/'` manually

realityking: That'd be implode: `implode('\', [$cacheDir, $spaceId])`

Unfortunately there's nothing like it that takes variadic arguments.

PR #83, opened 2017-04-03
Remove symfony/console from the dependencies

unfunco:
Symfony console is not used within the project and this being in the dependency manifest along with the version restriction means that this is not installable via Composer with Magento 2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,14 +6,13 @@
     "require": {
         "php": "^5.5.9|^7.0",
         "guzzlehttp/guzzle": "^6.2.1",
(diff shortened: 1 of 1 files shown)


realityking:
I indeed missed it in c352a787b0b2aa06dd28c035041dd3acc49bef14 but I'd reintroduce it anyway in #79.

Regarding Magento, a quick look into their tracker showed (magento/magento2#8738) they've already fixed the problem for an upcoming version.

That obviously doesn't help you right now. Until Magento has pushed a release with that fix your best bet might be to fork the SDK. 😞

PR #107, opened 2017-05-05
Add a HasSystemProperties trait and use it

### LINKED ISSUE #101: DeletedEntry should also have a getContentType() method
Currently `DeletedEntry` payloads look like this:

```json
{
  "sys": {
    "type": "DeletedEntry",
    "id": "someid",
    "space": {
      "sys": {
        "type": "Link",
        "linkType": "Space",
        "id": "somespace"
      }
    },
    "revision": 3,
    "createdAt": "2017-05-03T13:19:33.711Z",
    "updatedAt": "2017-05-03T13:19:33.711Z",
    "deletedAt": "2017-05-03T13:19:33.711Z",
    "contentType": {
      "sys": {
        "type": "Link",
        "linkType": "ContentType",
        "id": "theme"
      }
    }
  }
}
```

Yet the `DeletedEntry` class doesn't have the `getContentType()` method that e.g. `DynamicEntry` has. This makes it impossible to determine what type of thing was deleted, you have to parse the raw payload yourself to extract the data.

Are you accepting pull requests? There doesn't seem to be much activity here so I'm reluctant to post a solution myself, but if you acknowledge this then I will.

Jalle19:
Incidentally fixes #101

cc @realityking

### DIFF
--- a/src/Delivery/Asset.php
+++ b/src/Delivery/Asset.php
@@ -8,6 +8,8 @@ namespace Contentful\Delivery;
 
 class Asset extends LocalizedResource implements \JsonSerializable
 {
+    use HasSystemProperties;
+
     /**
      * @var array
      */
@@ -22,11 +24,6 @@ class Asset extends LocalizedResource implements \JsonSerializable
      * @var array
      */
     private $file;
-
-    /**
-     * @var SystemProperties
-     */
-    private $sys;
 
     /**
      * Asset constructor.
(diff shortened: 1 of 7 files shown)


realityking:
Thank you for the contribution. Adding methods like `Space::getContentType` that will always return `null` would make it harder to discover what data is available for any given entity. In the end we decided to go with a more conservative approach in #112.

PR #136, opened 2017-06-08
Fixes badly formatted error messages in the console commands.

### LINKED ISSUE #129: Remove <info> tag from messages in exceptions in console commands
The two commands in `\Contentful\Delivery\Console`, when throwing exceptions, format the error messages using <info> tags, but these are ignored by `symfony/console` as the errors are automatically formatted.

realityking:
The <info> tags can't be used when it's part of the exception message. Actually we don't even need esceptions here, we can just write our error message and terminate.

Fixes #129


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 * Retrieving a list of entries that contained multiple loops creates too many objects. **[BREAKING]** ([#105](https://github.com/contentful/contentful.php/pull/105))
   The new behavior is, that any entry that appears multiple times in the graph of the response will be the same instance.
 * The `contentful` script used to warm up/clear the cache was not marked as a binary in `composer.json` and thus not published to `vendor/bin`.
+* In console commands `<info>` can't be used as pat of an Exception message. ([#129](https://github.com/contentful/contentful.php/pull/129))
```
dborsatto: Typo _as pat of_

PR #138, opened 2017-06-09
Improve the way we create exceptions

realityking:
Adds a base class called ApiException for all API related exceptions and expose some additional information we can get from the response.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -172,33 +172,28 @@ abstract class Client
      */
     private function doRequest(RequestInterface $request, array $options)
     {
+        $exceptionMap = [
+            'InvalidQuery' => Exception\InvalidQueryException::class,
+            'AccessTokenInvalid' => Exception\AccessTokenInvalidException::class,
+            'NotFound' => Exception\NotFoundException::class,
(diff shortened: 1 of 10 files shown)


src/Client.php
```diff
@@ -172,33 +172,28 @@ protected function request($method, $path, array $options = [])
      */
     private function doRequest(RequestInterface $request, array $options)
     {
+        $exceptionMap = [
+            'InvalidQuery' => Exception\InvalidQueryException::class,
+            'AccessTokenInvalid' => Exception\AccessTokenInvalidException::class,
+            'NotFound' => Exception\NotFoundException::class,
+            'RateLimitExceeded' => Exception\RateLimitExceededException::class
+        ];
+
```
dborsatto: We could probably move this definition to _a better place_ ™. What do say about maybe having a static `ApiException::getExceptionMap()`?

realityking: Eventually we'll need to be able to override it to extend Exceptions for specific SDKs. I'd wanna wait to do so until we know what exactly we'll need.

PR #152, opened 2017-06-20
Adds DeletedContentType

dborsatto:


dborsatto:
I kind of want to check whether the sync API can secretly handle deleted content types as well, but they're simply not documented.

dborsatto:
Because this library will need some refactoring eventually, I've decided to be pragmatic and simply add `DeletedContentType` in the same place as `DeletedAsset`/`DeletedEntry`.

It's not the best place to put it, and by itself is kind of useless there, but as it's needed by other libraries, moving things around and deprecating classes feels like unnecessarily overcomplicating things. Because of this, pragmatism FTW 😅

PR #154, opened 2017-06-23
Adds the ability of setting the base URI on a per-request basis

dborsatto:
This is needed for the CMA `upload.contentful.com` endpoint.
Because of the nature of the CDA SDK, this is a change that is hard to test without using things like introspection and reflection, so for now I'd propose that the fact that we consider enough that all tests keep passing. The actual tests for this will be in the CMA SDK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 * Allow the exception map in `Client` to be overridden. This is done in preparation of the upcoming CMA SDK.
+* The third parameter `$options` of the `Client::request()` method now accepts an optional value with key `baseUri`. This is in preparation for the CMA SDK.
 
 ### Fixed
 * Slight fixes to error messages in exceptions thrown in the `ResourceBuilder`.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 * Allow the exception map in `Client` to be overridden. This is done in preparation of the upcoming CMA SDK.
+* The third parameter `$options` of the `Client::request()` method now accepts an optional value with key `baseUri`, which can be used to override the `Client`'s default internal value. Because of this, `Client::buildRequest()` now accepts an extra `$baseUri` optional parameter.
```
realityking: I'd write a less detailed changeling entry for this one as (hopefully) no user will ever interact with this method directly. You should also mention that this is prep work for the CMA.

dborsatto: `Client::request` is actually a protected method, so hopefully nobody actually interacts with it directly in any case. In this situation, I figured the changelog would be more for internal use...

realityking: I'd focus the Changelog on public use. Otherwise it's hard for users to tell what information is relevant for them or not.

PR #157, opened 2017-06-27
Adds handling of uploadFrom objects in the Preview API

dborsatto:


CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 * Allow the exception map in `Client` to be overridden. This is done in preparation of the upcoming CMA SDK.
 * The third parameter `$options` of the `Client::request()` method now accepts an optional value with key `baseUri`. This is in preparation for the CMA SDK.
-* Revamped the [reference documentation](https://contentful.github.io/contentful.php/api/) to be based on Sami and to include previous versions of the SDK. 
+* Revamped the [reference documentation](https://contentful.github.io/contentful.php/api/) to be based on Sami and to include previous versions of the SDK.
+* `UploadFromFile` now handles asset files which have been uploaded to `upload.contentful.com` but have yet to be processed. The class implements `Contentful\Delivery\File\FileInterface`, so it's not a breaking change (it actually fixes a possible edge case). This is also done in preparation for the upcoming CMA SDK.
```
realityking: I'd save the "not a breaking change" sentence, it sounds super defensive.

src/File/UploadFromFile.php
```diff
@@ -0,0 +1,76 @@
...
+     * @var string
+     */
+    private $contentType;
+
+    /**
+     * @var \JsonSerializable
+     */
+    private $uploadFrom;
+
+    public function __construct($fileName, $contentType, \JsonSerializable $uploadFrom)
```
realityking: Why is `$uploadFrom` an instance of `JsonSerializable` and not of `Link`?

src/File/UploadFromFile.php
```diff
@@ -0,0 +1,76 @@
+<?php
+/**
+ * @copyright 2015-2017 Contentful GmbH
+ * @license   MIT
+ */
+
+namespace Contentful\File;
+
+class UploadFromFile implements FileInterface
```
realityking: Would it make sense to add another interface `UploadableFileInterface` for this class and `UploadFile` so we can type hint for it in the CMA?

src/File/UploadFromFile.php
dborsatto: Because I'm not yet sure how to work with this in the CMA, so for the moment I'm only using `JsonSerializable`

src/File/UploadFromFile.php
dborsatto: I don't think we'll need that, to be honest...

realityking:
I thought about the naming over lunch. Here's what I came up with [drumroll]:

- `LocalUploadFile`
- `RemoteUploadFile`

WDYT?

dborsatto:
I like them:
- We should rename `UploadFrom` to `LocalUploadFile`...
- ...but we should also rename `UploadFile` to `RemoteUploadFile`, which we can't for BC reasons.

Ideas?

realityking:
For `RemoteUploadFile` this would work:

1.  Create `RemoteUploadFile` and move all code into it
2. Change `UploadFile` to extend `RemoteUploadFile` and mark it as `@deprecated`
3. CDA `ResourceBuilder` keeps building `UploadFile`
4. CMA `ResourceBuilder` uses the new `RemoteUploadFile`
5. In the next major release we change No. 3 and remove `UploadFile`

A bit of a hassle for such a rarely used class (that we also only introduced a few days ago) but the one extra file won't hurt us much.

src/File/UnprocessedFileInterface.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * @copyright 2015-2017 Contentful GmbH
+ * @license   MIT
+ */
+
+namespace Contentful\File;
+
+interface UnprocessedFileInterface extends \JsonSerializable
```
realityking: Should `UnprocessedFileInterface` extend `FileInterface`?

src/File/UploadFile.php
```diff
@@ -6,77 +6,9 @@
 
 namespace Contentful\File;
 
-class UploadFile implements FileInterface
+/**
+ * @deprecated As of version 2.1
```
realityking: According to [PSR-5](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#75-deprecated) this should look as follows:

```
@deprecated 2.1 Use RemoteUploadFile instead
@see RemoteUploadFile
```

src/File/UnprocessedFileInterface.php
dborsatto: I slightly like having them separate better, but I don't have any real preference towards either one 🤔

dborsatto:
@realityking philosophical question: do you think we should also change `Contentful\Tests\Unit\Delivery\UploadFileTest` to `Contentful\Tests\Unit\Delivery\RemoteUploadFileTest` (and change the actual class tested in the file as well)?

Issue #117, opened 2017-05-19
DynamicEntry get<FieldName>() returns different data structure in preview client vs published mode


riyunoa:
Just did a composer update today, and we noticed that the DynamicEntry get<FieldName>() function is now returning different data formats with the preview client than the normal client. Is there a reason for this?

e.g. in preview 'myField' = [en-US => [en-US => "this value"]
in normal mode 'myField' = [en-US => "this value"]

Is there a way we can get the data in the same format? Thanks!

realityking:
Hi @riyunoa! There's definitely something going wrong here.

Could you please let me know the following things:
* What version of the SDK are you using. (Will be shown if run `composer info contentful/contentful`
* Do you set a locale either when constructing the client or when making the request?

If you have a piece of code that shows the issue, that would be even better. Feel free to send it to support@contentful.com including credentials and mention it's in reference to this ticket and I'll take a look.

riyunoa:
Thanks for your response. We aren't setting the locale when constructing the client or making the request. We actually compared the DynamicEntry.php from the previous version (which returned the same response) and the current version's, and it looks like the `__call(...)` method has changed and is adding in the locale.

Composer details:

    $ composer info contentful/contentful
    name     : contentful/contentful
    descrip. : SDK for the Contentful Content Delivery API
    keywords :
    versions : * 1.2.0
    type     : library
    license  : MIT License (MIT) (OSI approved) https://spdx.org/licenses/MIT.html#licenseText
    source   : [git] https://github.com/contentful/contentful.php.git d3fd015dfe67c9dfdd744ae6531de7c1084d2e4f
    dist     : [zip] https://api.github.com/repos/contentful/contentful.php/zipball/d3fd015dfe67c9dfdd744ae6531de7c1084d2e4f d3fd015dfe67c9dfdd744ae6531de7c1084d2e4f
    names    : contentful/contentful

Constructing the clients:

    $this->contentfulClient = new Client(CmsService::CONTENTFUL_CONTENT_DELIVERY_TOKEN, CmsService::SPACE_ID);

    $this->contentfulPreviewClient = new Client(CmsService::CONTENTFUL_PREVIEW_TOKEN, CmsService::SPACE_ID, true);

Usage:

    if ($previewMode) {
        $content = $this->contentfulPreviewClient->getEntry(CmsService::HOMEPAGE_ENTRY_ID);
    } else {
[...]

PR #161, opened 2017-07-07
Rerecords the test interactions due to some changes to the space we use.

realityking:


dborsatto:
Is this all of them?

realityking:
Of course, I removed the entire directory and made completely new recordings.

realityking:
I reverted the updates to 
* `tests/fixtures/e2e_sync_preview.json`
* and `tests/fixtures/e2e_sync_preview_continue.json`

Otherwise they'd record the API with the fix for #117. Since that fix is currently flagged to only some orgs I wanna keep the recording for the "broken" behaviour.

dborsatto:
Ok 👍

PR #166, opened 2017-07-12
Configures the SDK for testing in API coverage environment

dborsatto:
It's mostly changes to the tests, except 2 things:
- `Contentful\Client` now has the method `getLogger` (which maybe could go in the changelog)
- `Contentful\Exception\ApiExeption` now uses `getHeaderLine` for accessing `X-Contentful-Request-Id`, which is safer as it returns an empty string if no header with that name is available.

The rest of the changes are in the `tests` directory, where basically I extracted all `Client` instantiations and moved them to the base `End2EndTestCase` class. As some of the tests involved cleaning up a cache directory, I simply moved these operations to the base `setUp` and `tearDown` methods.

Tests that are not suitable for running in a proxy environment are marked with the PHPUnit annotation `@requires Contentful`, and automatically skipped.

### DIFF
--- a/phpunit-api-coverage.xml
+++ b/phpunit-api-coverage.xml
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit
-        bootstrap="tests/bootstrap.php"
-        checkForUnintentionallyCoveredCode="true">
+<phpunit bootstrap="tests/bootstrap-api-coverage.php">
     <testsuites>
         <testsuite name="e2e">
             <directory>tests/E2E</directory>
@@ -13,4 +11,8 @@
             <directory suffix=".php">src</directory>
         </whitelist>
     </filter>
+
+    <php>
+        <env name="CONTENTFUL_CDA_SDK_TESTING_URL" value="http://localhost:5000/" />
+    </php>
 </phpunit>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit
-        bootstrap="tests/bootstrap.php">
+<phpunit bootstrap="tests/bootstrap.php">
     <testsuites>
         <testsuite name="unit">
             <directory>tests/Unit</directory>
--- a/src/Client.php
+++ b/src/Client.php
@@ -237,6 +237,14 @@ abstract class Client
     }
 
     /**
+     * @return LoggerInterface
+     */
+    public function getLogger()
+    {
+        return $this->logger;
+    }
+
+    /**
      * The name of the library to be used in the User-Agent header.
      *
      * @return string
--- a/tests/E2E/ErrorTest.php
+++ b/tests/E2E/ErrorTest.php
@@ -6,12 +6,11 @@
 
 namespace Contentful\Tests\E2E;
 
-use Contentful\Delivery\Client;
 use Contentful\Delivery\Query;
-
 use Contentful\Exception\RateLimitExceededException;
+use Contentful\Tests\Delivery\End2EndTestCase;
 
-class ErrorTest extends \PHPUnit_Framework_TestCase
+class ErrorTest extends End2EndTestCase
 {
     /**
      * @expectedException \Contentful\Exception\NotFoundException
@@ -19,7 +18,7 @@ class ErrorTest extends \PHPUnit_Framework_TestCase
      */
     public function testResourceNotFound()
     {
-        $client = new Client('b4c0n73n7fu1', 'cfexampleapi');
+        $client = $this->getClient('cfexampleapi');
 
         $client->getEntry('not-existing');
     }
@@ -30,7 +29,8 @@ class ErrorTest extends \PHPUnit_Framework_TestCase
      */
     public function testAccessTokenInvalid()
     {
-        $client = new Client('e5e8d4c5c122cf28fc1af3ff77d28bef78a3952957f15067bbc29f2f0dde0b50', 'cfexampleapi');
+        $client = $this->getClient('cfexampleapi_invalid');
+
         $client->getEntries();
     }
 
@@ -40,7 +40,8 @@ class ErrorTest extends \PHPUnit_Framework_TestCase
      */
     public function testInvalidQuery()
     {
-        $client = new Client('b4c0n73n7fu1', 'cfexampleapi');
+        $client = $this->getClient('cfexampleapi');
+
         $query = (new Query())
             ->where('name', 0);
 
@@ -48,12 +49,14 @@ class ErrorTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @requires API no-coverage-proxy
      * @expectedException \Contentful\Exception\RateLimitExceededException
      * @vcr e2e_error_rate_limit.json
      */
     public function testRateLimitExceeded()
     {
-        $client = new Client('8740056d546471e0640d189615470cc12ce2d3188332352ecfb53edac59c4963', 'bc32cj3kyfet', true);
+        $client = $this->getClient('bc32cj3kyfet_preview');
+
         $query = new Query();
 
         try {
--- a/tests/E2E/GzipEncodingTest.php
+++ b/tests/E2E/GzipEncodingTest.php
@@ -6,21 +6,19 @@
 
 namespace Contentful\Tests\E2E;
 
-use Contentful\Delivery\Client;
-use Contentful\Log\ArrayLogger;
 use Contentful\Delivery\Query;
+use Contentful\Tests\Delivery\End2EndTestCase;
 
-class GzipEncodingTest extends \PHPUnit_Framework_TestCase
+class GzipEncodingTest extends End2EndTestCase
 {
     /**
+     * @requires API no-coverage-proxy
      * @vcr e2e_gzip_encoding.json
      */
     public function testContentEncodingHeader()
     {
-        $logger = new ArrayLogger;
-        $client = new Client('b4c0n73n7fu1', 'cfexampleapi', false, null, [
-            'logger' => $logger
-        ]);
+        $client = $this->getClient('cfexampleapi_logger');
+        $logger = $client->getLogger();
 
         $query = (new Query())
             ->setLocale('*');
@@ -30,7 +28,7 @@ class GzipEncodingTest extends \PHPUnit_Framework_TestCase
 
         $this->assertEquals('gzip', $logEntry->getRequest()->getHeaderLine('Accept-Encoding'));
 
-        // Need to check 'x-encoded-content-encoding' as curl is automatically decompressing the response
-        $this->assertEquals('gzip', $logEntry->getResponse()->getHeaderLine('x-encoded-content-encoding'));
+        // Need to check 'X-Encoded-Content-Encoding' as curl is automatically decompressing the response
+        $this->assertEquals('gzip', $logEntry->getResponse()->getHeaderLine('X-Encoded-Content-Encoding'));
     }
 }
--- a/tests/End2EndTestCase.php
+++ b/tests/End2EndTestCase.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * @copyright 2015-2017 Contentful GmbH
+ * @license   MIT
+ */
+
+namespace Contentful\Tests\Delivery;
+
+use Contentful\Delivery\Client;
+use Contentful\Log\ArrayLogger;
+
+class End2EndTestCase extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var string
+     */
+    protected static $cacheDir = '';
+
+    public static function setUpBeforeClass()
+    {
+        self::$cacheDir = __DIR__ . '/../build/cache';
+    }
+
+    /**
+     * @param string $key
+     *
+     * @return Client
+     */
+    protected function getClient($key)
+    {
+        $testingUrl = getenv('CONTENTFUL_CDA_SDK_TESTING_URL');
+        $options = $testingUrl
+            ? ['uriOverride' => $testingUrl]
+            : [];
+
+        switch ($key) {
+            case 'cfexampleapi':
+                return new Client('b4c0n73n7fu1', 'cfexampleapi', false, null, $options);
+            case 'cfexampleapi_preview':
+                return new Client('e5e8d4c5c122cf28fc1af3ff77d28bef78a3952957f15067bbc29f2f0dde0b50', 'cfexampleapi', true, null, $options);
+            case 'cfexampleapi_cache':
+                return new Client('b4c0n73n7fu1', 'cfexampleapi', false, null, array_merge($options, ['cacheDir' => self::$cacheDir]));
+            case 'cfexampleapi_logger':
+                return new Client('b4c0n73n7fu1', 'cfexampleapi', false, null, array_merge($options, ['logger' => new ArrayLogger()]));
+            case 'cfexampleapi_tlh':
+                return new Client('b4c0n73n7fu1', 'cfexampleapi', false, 'tlh', $options);
+            case 'cfexampleapi_invalid':
+                return new Client('e5e8d4c5c122cf28fc1af3ff77d28bef78a3952957f15067bbc29f2f0dde0b50', 'cfexampleapi', false, null, $options);
+            case '88dyiqcr7go8':
+                return new Client('668efbfd9e398181166dec5df5a500aded96dbca2916646a3c7ec37082a7b756', '88dyiqcr7go8', false, null, $options);
+            case '88dyiqcr7go8_preview':
+                return new Client('81c469d7241ca02349388602dfc14107157063a6901c378a56e1835d688970bf', '88dyiqcr7go8', true, null, $options);
+            case 'bc32cj3kyfet_preview':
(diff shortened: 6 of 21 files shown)


dborsatto:
I'm not crazy about creating all those instances in `End2EndTestCase::setUpBeforeClass` instead of lazy-loading them, but to be honest most of the times when the test suite is executed, all those client instances are actually needed, so that's probably a reasonable approach.

src/Exception/ApiException.php
```diff
@@ -50,7 +50,7 @@ public function __construct(GuzzleRequestException $previous, $message = '')
 
         if ($this->previous->hasResponse()) {
             $this->response = $this->previous->getResponse();
-            $this->requestId = $this->response->getHeader('X-Contentful-Request-Id')[0];
+            $this->requestId = $this->response->getHeaderLine('X-Contentful-Request-Id');
```
realityking: What's the issue you ran into?

From the PSR-7 spec:
> Use the `getHeaderLine()` method to retrieve a header value as a string containing all header values of a case-insensitive header by name concatenated with a comma.

I'm not sure that's why we want.

phpunit-api-coverage.xml
```diff
@@ -13,4 +11,8 @@
             <directory suffix=".php">src</directory>
         </whitelist>
     </filter>
+
+    <php>
+        <const name="SDK_TESTING_ENV" value="api_coverage" />
```
realityking: Probably just a question of style but did you consider using an env variable instead?

tests/E2E/GzipEncodingTest.php
```diff
@@ -30,7 +28,7 @@ public function testContentEncodingHeader()
 
         $this->assertEquals('gzip', $logEntry->getRequest()->getHeaderLine('Accept-Encoding'));
 
-        // Need to check 'x-encoded-content-encoding' as curl is automatically decompressing the response
-        $this->assertEquals('gzip', $logEntry->getResponse()->getHeaderLine('x-encoded-content-encoding'));
+        // Need to check 'X-Encoded-Content-Encoding' as curl is automatically decompressing the response
+        $this->assertEquals('gzip', $logEntry->getResponse()->getHeaderLine('X-Encoded-Content-Encoding'));
```
realityking: Is this just a stylistic change? Because either should work.

tests/bootstrap-api-coverage.php
```diff
@@ -1,7 +1,11 @@
 <?php
+
 /**
- * @copyright 2015-2016 Contentful GmbH
+ * This file is part of the contentful.php package.
```
realityking: Personal preference but I'd not mix this within one project :/

tests/End2EndTestCase.php
```diff
@@ -0,0 +1,82 @@
...
+    const CACHE_DIR = __DIR__ . '/../build/cache';
+
+    /**
+     * @var Client[]
+     */
+    private static $clients = [];
+
+    public static function setUpBeforeClass()
+    {
+        self::$clients = [
```
realityking: What's the reason for the eager creation? Intuitively I'd have implemented this the same as the CMA, it's easier to reason about.

tests/E2E/ErrorTest.php
```diff
@@ -40,20 +40,23 @@ public function testAccessTokenInvalid()
...
+        $client = $this->getClient('cfexampleapi');
+
         $query = (new Query())
             ->where('name', 0);
 
         $client->getEntries($query);
     }
 
     /**
+     * @requires Contentful
```
realityking: Clever! Let me think about that one for a bit 😄

src/Client.php
```diff
@@ -243,6 +243,14 @@ protected function getExceptionMap()
     }
 
     /**
+     * @return LoggerInterface
+     */
+    public function getLogger()
```
realityking: This should definitely go into the changelog.

tests/End2EndTestCase.php
```diff
@@ -0,0 +1,82 @@
...
+            'cfexampleapi_logger' => new Client('b4c0n73n7fu1', 'cfexampleapi', false, null, ['logger' => new ArrayLogger()]),
+            'cfexampleapi_tlh' => new Client('b4c0n73n7fu1', 'cfexampleapi', false, 'tlh'),
+            'cfexampleapi_invalid' => new Client('e5e8d4c5c122cf28fc1af3ff77d28bef78a3952957f15067bbc29f2f0dde0b50', 'cfexampleapi'),
+            '88dyiqcr7go8' => new Client('668efbfd9e398181166dec5df5a500aded96dbca2916646a3c7ec37082a7b756', '88dyiqcr7go8'),
+            '88dyiqcr7go8_preview' => new Client('81c469d7241ca02349388602dfc14107157063a6901c378a56e1835d688970bf', '88dyiqcr7go8', true),
+            'bc32cj3kyfet_preview' => new Client('8740056d546471e0640d189615470cc12ce2d3188332352ecfb53edac59c4963', 'bc32cj3kyfet', true),
+        ];
+    }
+
+    public function setUp()
```
realityking: This one I'm not a fan of in the parent class. That code will run before/after every single test method, whether it uses the cache or not.

src/Exception/ApiException.php
dborsatto: The coverage tools strips out all headers by default, so this raised a warning because that header is not available.

PSR-7 assumes that headers can have multiple values, and for some it _can_ be true (like set-cookie), but the vast majority is basically a key => value scenario. Because of this, `getHeaderLine` is simply a shortcut instead of doing something like
``` php
$requestIdHeader = $this->response->getHeader('X-Contentful-Request-Id');
$this->requestId = $requestIdHeader ? $requestIdHeader[0] : '';
```

`X-Contentful-Request-Id` will only ever be a single line, so there's really no problem using `getHeaderLine`, and on the plus side we don't have to make an explicit check ourselves.

tests/E2E/GzipEncodingTest.php
dborsatto: Yes, Guzzle lowercases everything, so this is just for consistency with routine use.

tests/End2EndTestCase.php
dborsatto: The CMA client only needs a token, so that's why that approach works there. Here we need a bunch of different clients with very different requirements, so whereas the CMA gets away with `getReadOnlySpaceManager`/`getReadWriteSpaceManager`, here we'd need 9 different getter methods, or one with a huge switch/case, where the client would be instantiated on-demand and then saved.

But as the test suite is most often execute as a whole, it actually make sense to have the clients defined like this (instead of created/destroyed multiple times), and I could actually improve on this by checking that if the array is not empty, then returning early because clients are already set up.

tests/E2E/ErrorTest.php
dborsatto: PHPUnit has this builtin with something like the PHP version or the existence of certain functions, but it was actually easy to extend. I find it nice and declarative.

tests/End2EndTestCase.php
dborsatto: Yeah, I was trying to make something smarter than this, but in the end the performance impact is basically negligible...

Maybe I could implement this using an annotation like I did with `@requires`, something like `@uses-cache`?

tests/bootstrap-api-coverage.php
dborsatto: I'd like to avoid what I did in the CMA (with a single commit changing everything) but instead changing the header whenever the file is being changed...
This also needed updating, as it was saying 2015-2016, so it was two birds with one stone.

phpunit-api-coverage.xml
dborsatto: I didn't actually think about it, to be honest.
Do you mean simply changing this to use `<env>` instead of `<const>` ([docs](https://phpunit.de/manual/current/en/appendixes.configuration.html#appendixes.configuration.php-ini-constants-variables)), or actually having environment variables set in whatever thing is running the coverage tool?

realityking: The former, using `<env>`.

My logic would be that environment variables were designed for these things. Having phpunit generate constants trips up static analysis, for example PhpStorm will claim that `SDK_TESTING_ENV` is an undefined constant.

src/Exception/ApiException.php
realityking: My concern were more around what happens if we have a bug and duplicate a header. (has happened before)

Of course with the current code we have the opposite problem that if we have a bug that the header isn't set if tails 🤔

tests/End2EndTestCase.php
realityking: I think this is overkill. The whole cache thing is only used in a single class, why move the handling for it to the base class?

tests/bootstrap-api-coverage.php
realityking: Personally I'd rather see the big change then this. Otherwise this repo will be inconsistent for years...

tests/End2EndTestCase.php
realityking: The big downside of this approach is that it breaks the isolation between test cases. This is most obvious in the `cfexampleapi_logger` case where the logger is shared across all tests using it. (It's only one right now)

Let me think about this one a bit too :)

phpunit-api-coverage.xml
dborsatto: Yeah, Atom does the same. I'll change this 👍

tests/End2EndTestCase.php
dborsatto: To be honest I don't remember why I started moving the cache management to the base class 🤔

src/Exception/ApiException.php
dborsatto: > My concern were more around what happens if we have a bug and duplicate a header. (has happened before)

What bug?

dborsatto:
@realityking I've updated the way Client objects are created, please take a look.

src/Exception/ApiException.php
realityking: I mean if our API has a bug that a header gets sends twice. Similar things have happened to us.

I think neither implementation is 100% robust for corner cases. We need something like `getFirstHeaderLine()`.

tests/End2EndTestCase.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * @param string $key
+     *
+     * @return Client
+     */
+    protected function getClient($key)
+    {
+        $options = getenv('CONTENTFUL_CDA_SDK_TESTING_URL')
```
realityking: Stylistic but IMHO it's nicer to only call `getenv` once and then check/use the variable.

tests/End2EndTestCase.php
```diff
@@ -0,0 +1,79 @@
...
+            ? ['uriOverride' => getenv('CONTENTFUL_CDA_SDK_TESTING_URL')]
+            : [];
+
+        switch ($key) {
```
realityking: This looks much better. :) Any particular reason you went for a switch statement instead of many methods?

One advantage of methods would be that we could parameterize them which would cut down on the number of different methods/cases we need and we could inject things like the logger or cache path at that point.

Though overall not a big difference.

src/Exception/ApiException.php
dborsatto: To be honest, making some weird workarounds for possible bugs that may not ever happen for a very specific implementation where the worst thing that could happen is to actually report the ID twice (should such a bug ever arise), well, it seems a bit silly to me...

tests/E2E/ErrorTest.php
realityking: While I'm not the biggest annotations fan, this seems fine.

What I don't think is good is the name. All E2E tests require Contentful, what makes these different?

tests/End2EndTestCase.php
```diff
@@ -0,0 +1,79 @@
...
+
+        foreach (['class', 'method'] as $depth) {
+            if (empty($annotations[$depth]['requires'])) {
+                continue;
+            }
+
+            $requires = array_flip($annotations[$depth]['requires']);
+
+            if (isset($requires['Contentful'])) {
+                return $this->markTestSkipped('Tests against Preview API are not enabled with this configuration.');
```
realityking: I think this is not 100% accurate, some of the skipped tests don't use the Preview API. (e.g. the URI override one)

tests/E2E/ErrorTest.php
dborsatto: Because they require to speak _directly_ to Contentful, and not a proxy of some kind. I can change this to `@requires Contentful no-proxy`, if you think it looks better

src/Exception/ApiException.php
realityking: 🤷‍♂️ Once we go down the path of hardening this might as well do it right

I don't feel too strongly about this. We use `getHeader()` all over the place (check the RateLimitException classes in both SDKs for others), it might be nice to just roll a helper function somewhere that takes care of this.

tests/End2EndTestCase.php
dborsatto: Yeah, I wrote that message before the URI override one was added to the list, I'll rephrase this

tests/E2E/ErrorTest.php
realityking: Oh wow, I didn't guess that meaning at all.

How about something like `@requires No coverage proxy`? That makes it clear we're not talking about normal HTTP proxies (which should work)

src/Exception/ApiException.php
dborsatto: Honestly I'd rather patch the coverage tool to let this header pass through (which would make a lot of sense anyway) than going down the path of fixing possible header issues (and reverting this change), what do you think?

realityking: Patching the coverage is probably a good idea either way. Again I don't feel to strongly whether this line looks one way or another but if we change it, we should change it everywhere.

tests/E2E/ErrorTest.php
dborsatto: Standard `@requires` annotations have a format `@requires SUBJECT ATTRIBUTE`, such as `@requires PHP 7.0` or `@requires extension ext-soap`. I suggest we use a similar structure for consistency, maybe `@requires API no-proxy`

tests/End2EndTestCase.php
dborsatto: No real difference but sure 👍

tests/E2E/ErrorTest.php
realityking: Fair enough. Just trying to make it clear we're not talking about normal HTTP proxies. (It'd be fun to rewrite the coverage tool being an actual proxy!)

`@requires API no-coverage-proxy`?

tests/End2EndTestCase.php
dborsatto: Mostly because not all methods would have meaningful names.
For instance `cfexampleapi` and `88dyiqcr7go8` (and their respective `_preview` versions) are for all intents and purposes the same, but the things they need for testing are in two different spaces.

Some of them would be hard to describe using a name, so we might as well not do that and just use a key...

tests/E2E/ErrorTest.php
dborsatto: Sounds good 👍

src/Exception/ApiException.php
dborsatto: Because doing that would be out of scope, I'll patch the tool and postpone this discussion 🙂

tests/End2EndTestCase.php
realityking: Now that you mention it, that might actually be a problem. I can't tell anymore which space is the one used for rate limit tests for example.

But let's keep it like this for now, can always refactor it later.

src/Client.php
realityking: Based on the discussion downstream. Do you think this has merits on its own or only for testing? If it's the latter I suggest we refactor `getClient()` a tiny bit.

tests/End2EndTestCase.php
dborsatto: Well, same as before, you can look at the `ErrorTest.php` and see the space being used, it doesn't really change that 🙂

src/Client.php
dborsatto: I feel that's actually something worth doing either way.
For instance, this makes what I wanted to do in https://github.com/contentful/contentful.php/pull/162 much easier, as you don't need to keep the logger around because you just can retrieve it from the client.

The fact that it was added to cover a specific need here doesn't mean it'd not be useful overall, on the contrary...

realityking: 👍 

Though I disagree that this would be any help for #162 - the default `NullLogger` won't do you any good.

dborsatto: Yeah, but it would be a matter of ease of use.
Maybe you're setting up the client somewhere, and using it somewhere else. This way you don't have to keep yourself the logger around, you can simply rely on the client to store it for you...

dborsatto:
I've updated:
- Annotations (now `@requires API no-coverage-proxy`)
- Reverted `getHeaderLine` in `ApiException`
- Changed the skipped test error message
- Stored the result of `getenv` in a local variable

I think I've address all concerns, am I missing something?

phpunit.xml.dist
```diff
@@ -28,4 +27,5 @@
             <directory suffix=".php">src</directory>
         </whitelist>
     </filter>
+
```
realityking: This looks like an accident ;)

dborsatto: I had added the env config to that file as well, but later I've removed it, that's why there a leftover space there. I'll remove it 👍

PR #171, opened 2017-07-25
Removes static helpers in favor of simple functions

dborsatto:
This PR removes the use of helpers `DateHelper` and `JsonHelper`, which will be deprecated in v2.2 and removed in v3.

JsonHelper is replaced by corresponding functions already provided by Guzzle, which do a very similar thing (wrapping the default `json_*` functions and throwing exception in case of failure). For compatibility reason, JsonHelper still catches Guzzle's exceptions (InvalidArgumentException) and then re-throws another type (RuntimeException), which is the same as before, so nothing should actually change for users relying on these methods.
`DateHelper:: formatForJson` is replaced by a simple function defined in `src/functions.php`, with the same signature.

### DIFF
--- a/src/DateHelper.php
+++ b/src/DateHelper.php
@@ -12,19 +12,16 @@ class DateHelper
      * Unfortunately PHP has no easy way to create a nice, ISO 8601 formatted date string with milliseconds and Z
      * as the time zone specifier. Thus this hack.
      *
-     * @param  \DateTimeImmutable $dt
+     * @param  \DateTimeImmutable $date
      *
      * @return string ISO 8601 formatted date
+     *
+     * @deprecated 2.2 Use Contentful\format_date_for_json($date) instead
+     *
+     * @see Contentful\format_date_for_json()
      */
-    public static function formatForJson(\DateTimeImmutable $dt)
+    public static function formatForJson(\DateTimeImmutable $date)
     {
-        $dt = $dt->setTimezone(new \DateTimeZone('Etc/UTC'));
-        $result = $dt->format('Y-m-d\TH:i:s') ;
-        $milliseconds =floor($dt->format('u')/1000);
-        if ($milliseconds > 0) {
-            $result .= '.' . str_pad($milliseconds, 3, '0', STR_PAD_LEFT);
-        }
-
-        return $result . 'Z';
+        return format_date_for_json($date);
     }
 }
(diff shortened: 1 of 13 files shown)


src/DateHelper.php
```diff
@@ -12,19 +12,16 @@ class DateHelper
      * Unfortunately PHP has no easy way to create a nice, ISO 8601 formatted date string with milliseconds and Z
      * as the time zone specifier. Thus this hack.
      *
-     * @param  \DateTimeImmutable $dt
+     * @param  \DateTimeImmutable $date
```
loudmouth: hurray for readable variable names! 👏

src/functions.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+  * Unfortunately PHP has no easy way to create a nice, ISO 8601 formatted date string with milliseconds and Z
+  * as the time zone specifier. Thus this hack.
+  *
+  * @param \DateTimeImmutable $date
+  *
+  * @return string ISO 8601 formatted date
+  */
+ function format_date_for_json(\DateTimeImmutable $date)
```
loudmouth: i'm a bit confused by this change. the code looks the same as the old method, but there is also added inconsistency in that there is mixed snake_case and camelCase in the SDK now.

loudmouth: What was the impetus for deprecating the original method if the underlying logic is the same?

dborsatto: About the inconsistency in naming, that's a general convention in PHP, using `snake_case` for simple functions and `camelCase` when they are class methods 🙂 

About the reasons of the change, I'm generally against having static helper classes, mainly because in my experience that never ends well. Overall I try to have classes represent _something_, rather than being ways to group together functions: that's what namespaces are for. Unfortunately supporting PHP 5.5 means I can't add nice `use function Contentful\format_date_for_json;` statements at the beginning like I'm doing [here](https://github.com/contentful/contentful-management.php/pull/47/commits/cc1ac0233a30640266be50a443fd6baf2686d8d0#diff-f8505abedeb59d0333424cdf94cdeabeR12), but that's the nature of the beast.

As you can see, I'm using `GuzzleHttp\json_encode` and `GuzzleHttp\json_decode`, which follow the same logic. Overall it's probably a matter of preferences, though I can't shake the feeling that if I'm creating _a class around a static function (and this function does not access anything inside that class)_, then this function should be better off as standalone 🤔

At least this is for PHP, where OO and procedural code coexist _peacefully_ 😆

loudmouth: Awesome explanation. Thanks :-)

dborsatto: 🙇

Issue #174, opened 2017-09-25
Serialization of exception which contains closures in stacktrace cause problems


bestit-tkellner:
The serialize function of src/Log/LogEntry.php file cause problems if the exception contains closures in stack trace because it's not possible to serialize this.

dborsatto:
I'll try to reproduce the bug, but your description makes complete sense.
The problem is, I'm not completely sure how to work around this. I'll need to investigate a bit 🤔

bestit-tkellner:
I've created a pr with a fix for this problem. I use the traceAsString function of the exception class to prevent this problem.

PR #175, opened 2017-09-25
#174 Modify exception to prevent problems with closures in stack trace.

bestit-tkellner:
This pr fixed the problem described in issue #174

### DIFF
--- a/src/Log/LogEntry.php
+++ b/src/Log/LogEntry.php
@@ -58,6 +58,43 @@ class LogEntry implements \Serializable
         $this->exception = $exception;
         $this->duration = $duration;
         $this->response = $response;
+
+        if ($exception instanceof \Exception) {
+            $this->modifyTrace();
+        }
+    }
+
+    /**
+     * Modify closures in exception stack trace, because they aren't serializable.
+     */
+    private function modifyTrace()
+    {
+        $exception = $this->getException();
+
+        $traceProperty = (new \ReflectionClass('Exception'))->getProperty('trace');
+        $traceProperty->setAccessible(true);
+
+        do {
+            $trace = $traceProperty->getValue($exception);
+
+            array_walk_recursive($trace, function (&$value) {
+                if ($value instanceof \Closure) {
+                    $closureReflection = new \ReflectionFunction($value);
+
+                    $value = sprintf(
+                        '(Closure in file %s at line %s)',
+                        $closureReflection->getFileName(),
+                        $closureReflection->getStartLine()
+                    );
+                } elseif (is_object($value)) {
+                    $value = sprintf('object(%s)', get_class($value));
+                } elseif (is_resource($value)) {
+                    $value = sprintf('resource(%s)', get_resource_type($value));
+                }
+            });
+
+            $traceProperty->setValue($exception, $trace);
+        } while (($exception = $exception->getPrevious()) instanceof \Exception);
     }
 
     /**
--- a/tests/Unit/Log/LogEntryTest.php
+++ b/tests/Unit/Log/LogEntryTest.php
@@ -35,6 +35,44 @@ class LogEntryTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals($entry->getRequest()->getUri(), $serialized->getRequest()->getUri());
     }
 
+    public function testSerializeWithException()
+    {
+        $errorMessage = 'error messages';
+        $errorCode = 999;
+        $exception = null;
+
+        $closure1 = function ($closure) {
+            $closure();
+        };
+
+        $closure2 = function () use (&$exception, $errorMessage, $errorCode) {
+            $exception = new \Exception($errorMessage, $errorCode);
+        };
+
+        $closure1($closure2);
+
+        /** @var \Exception $exception $traceBeforeFix */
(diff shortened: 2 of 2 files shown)


dborsatto:
Hey @bestit-tkellner,

the problem with this solution is that it is not bidirectional. In fact, the `unserialize` method should be updated as well, otherwise the result on the process with be an array, and not an `Exception` object.

In order for me to approve this, the `unserialize` method must be updated taking into account 3 scenarios:

- `$data->exception` is `null`, easy peasy.
- `$data->exception` is an instance of `Exception`, as it was serialized using a previous version of the SDK.
- `$data->exception` is an array, in which case it should be manually converted to an `Exception` object. The problem with this approach is, you can inject `code` and `message` through the constructor, `file` and `line` would be trickier but possible (you'd probably have to resort to Reflection), but I'm not sure what could be done about `trace`. I'm looking at the docs, but I'm not sure if there's a clean way to do this _and_ not to lose some info.

bestit-tkellner:
We have problems in two projects because of this. So i hope this pr can fix this problem fast. ;-)

dborsatto:
I can't 100% promise that I'll make it, but I'll definitely try to work on this tomorrow!

dborsatto:
@bestit-tkellner ok so, I check your PR locally, and overall I'm good with it, I'd just have a couple changes which are included in [this gist](https://gist.github.com/dborsatto/96cffc58c6efcd5d6f72d791ac5ed8bb).

Basically, I've added a couple more tests to verify that the trace was indeed changed (although the fact that _the test passes_ and it doesn't raise an error kind of implies that).

I also made a couple of changes to `modifyTrace`. Using `array_walk_recursive` we can actually get rid of the outer `array_walk`. Iterating on a few more scalar values shouldn't be a problem, and it fixes having two nested closures, both with by-reference parameters, something which I'm not really crazy about. If you still prefer to optimize by applying `array_walk_recursive` only on the `args` array, then I ask you to please use a `foreach` instead of `array_walk`.

Finally, it's a small formatting preference, but I've turned

``` php
array_walk_recursive(
    $param,
    function (&$value) {
        // Something
    }
);
```

into

``` php
array_walk_recursive($param, function (&$value) {
    // Something
});
```
It's a personal preference, but it's shorter and indentation doesn't get as nested 🙂 

I went back and forth between applying `modifyTrace` in the constructor versus in the `serialize` method. It could be a rather demanding operation (reflection can be slow), so using it on-demand would be better for performance. However conceptually I think that `serialize` shouldn't modify the inner object, so we could create a copy of the exception, modify that copy, and include it in the serialized object, but this would mean that `$a !== unserialize(serialize($a))`. Eventually, keeping `modifyTrace` in the constructor guarantees the most consistency, and hopefully `$exception` should be `null` most of the times, so this shouldn't impact performance too much.

I would love if you could update the PR including my suggestions. If you do so, I'll merge it, then update the changelog and release a new version by the end of the day or tomorrow at the latest 👍

bestit-tkellner:
Hi @dborsatto,
thanks a lot! I agree with your changes (thumbsup).

I think it's a good idea to remove the additional array_walk and not to increase the complexity.

Formatting, your choice ;-)

At last i think it's the better way to to this in the constructor.

dborsatto:
Fixes #174

dborsatto:
@bestit-tkellner I just released `2.2.0`, I had to release a minor version and not a simple bugfix because there were a couple deprecations introduced in previous commits, and however small, I didn't want to introduce them in a bugfix release.

Let me know if your problems have been solved! It's been great having you contribute, I hope to see PRs from you again in the future 🙂

bestit-tkellner:
Unfortunately not :-(

I've just seen that we use version 1.0.0 ... And it's not possible to upgrade easily to version 2.2 

It's possible that you can release this bugfix as 1.0.1 too?

dborsatto:
Hey @bestit-tkellner,

unfortunately that's not something that I can do. The repo has seen a lot of development since version `1.0`, and the way it's set up doesn't allow for easy backporting. While I don't exclude restructuring it in the future (to have version-based branches of some sort), backporting this fix to `1.0` (which is also the oldest version in the v1 series) would mean that for consistency I'd have to backport _all_ other fixes that have been implemented since (and as you can see in the [releases page](https://github.com/contentful/contentful.php/releases), it's quite a few things).

While it's true that there have been breaking changes between version `1.x` and version `2`, I think that they should be sufficiently self-contained to be able to upgrade with minimal effort. In fact, taking a look at the [release notes](https://github.com/contentful/contentful.php/releases/tag/2.0.0), I'd say that the only thing that you _really_ should pay attention to is this part:

```
* Moved file classes to a sub-namespace Contentful\File **[BREAKING]**.
  * Contentful\File to Contentful\File\File
  * Contentful\ImageFile to Contentful\File\ImageFile
  * Contentful\ImageOptions to Contentful\File\ImageOptions
```

But this is something that could be automated by a global search/replace. The rest of the changes are rather internal, or cover a minimal percentage of actual use cases.

I'm sorry to say I can't do what you ask, but hopefully you can understand how doing that would result in a lot of work for am extremely specific use case, something which I'm afraid I can't justify doing 😞

PR #176, opened 2017-11-02
Add an easier way to sync

slootjes:
While playing around with the Sync Manager I thought it would be a better developer experience to just have 1 loop for syncing instead of calling multiple methods. To get all content:

```
$syncManager = $client->getSynchronizationManager();
foreach ($syncManager->sync() as $result) {
    // use the $result object how you like
}
```
Or pass a previously stored token:

```
$syncManager = $client->getSynchronizationManager();

$token = '<token from a previous sync>';
foreach ($syncManager->sync($token) as $result) {
   // use the $result object how you like
}
```

Pass a query to get specific content:
```
$syncManager = $client->getSynchronizationManager();

$query = new \Contentful\Delivery\Synchronization\Query;

foreach ($syncManager->sync(null, $query) as $result) {
    // use the $result object how you like
}
```

Because running/adding the tests did not work out of the box I added a dev requirement for phpunit/phpunit so it can be tested standalone with `php vendor/phpunit/phpunit/phpunit`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,7 @@
     "require-dev": {
         "php-vcr/phpunit-testlistener-vcr": "^2.0",
         "jakub-onderka/php-parallel-lint": "^0.9.2",
-        "jakub-onderka/php-console-highlighter": "^0.3.2",
-        "phpunit/phpunit": "~4.0"
+        "jakub-onderka/php-console-highlighter": "^0.3.2"
     },
     "autoload": {
         "files": ["src/functions.php"],
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     "require-dev": {
         "php-vcr/phpunit-testlistener-vcr": "^2.0",
         "jakub-onderka/php-parallel-lint": "^0.9.2",
-        "jakub-onderka/php-console-highlighter": "^0.3.2"
+        "jakub-onderka/php-console-highlighter": "^0.3.2",
+        "phpunit/phpunit": "~4.0"
     },
     "autoload": {
         "files": ["src/functions.php"],
--- a/src/Delivery/Synchronization/Manager.php
+++ b/src/Delivery/Synchronization/Manager.php
@@ -54,6 +54,31 @@ class Manager
     }
 
     /**
+     * @param null $token
+     * @param Query|null $query
+     *
+     * @return \Generator|void
+     */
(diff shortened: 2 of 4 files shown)


slootjes:
I don't think codeclimate understands the concept of generators, doesn't seem like something I can fix.

dborsatto:
Hey @slootjes,
sorry for replying just now, it's been a very busy few weeks.

Admittedly I'll have to test this and understand how it works. I replaced the person who built the SDK a few months ago, and the Sync API part is something I haven't really touched yet--just being honest 🙂 

A couple of things, though:
- Codeclimate is complaining about the fact that you're using an ELSE statement. In that case, you can easily refactor that bit by using a ternary condition.
- Tests are run by using a `.phar` version of PHPUnit, you can get it [here](https://phar.phpunit.de/phpunit-5.7.phar). At the moment, testing is done using either PHPUnit 4.8 or 5.7 as Phars, but in the future I'd like to move to a dev dependency on the `v6` branch.

I'll try to take a look at this as soon as possible. In the meanwhile, thanks for your contribution!

slootjes:
Thanks for your reply. I've changed the if/else to a ternary operator and removed the phpunit dependency. Not sure why SensioLabsInsight isn't updating...

slootjes:
Just a bump to say that SensioLabsInsight never reported back... I don't expect any issues with it though.

dborsatto:
Hey, I'm currently on holiday right now, I just wanted to drop a comment to tell you that I saw this, and I'll get to it once I'm back ☺️

dborsatto:
Hey @slootjes,

I was finally able to get to this! And I must say I really like it, using generators instead of regular arrays is a pretty smart choice 🎉 

I have some minor nitpicks, but it's things I'll fix myself after merging the PR:
* The `@return` of `sync` should only be `\Generator`, because technically that's what it is. It'd be cool to have some sort of generics in PHP so we could say something like `\Generator<Result>`, but unfortunately we're not there yet.
* I'll add an entry to the changelog to document the new feature.
* I'll rename test method to `testSyncFull` and the fixture to `e2e_sync_full.json` to better reflect the meaning of the test.

Because there have been some other changes recently, I'll merge the PR as is, which will possible break a few things on the master branch, which I'll immediately address anyway 🙂

PR #181, opened 2018-01-10
PSR-6 integration

magnusnordlander:
Hi!

We find ourselves needing more flexibility in the cache generation (specifically, we're using a separate stage space, and we're deploying on Heroku, so ideally we'd like to be able to use the same slug that we build for the stage environment as we do for the prod environment).

This PR replaces the Filesystem with a PSR-6 cache, and pulls in the `symfony/cache` component to provide cache adapters for a Filesystem and a Null cache (ensuring that there are no API changes).

I saw the now closed PSR-16 PR, and made sure to steer clear of the issues identified within. I used PSR-6 seeing as that is the more capable of the two interfaces, but I'm not adamant about it. 

This PR should be mergable from our standpoint, but I do have a couple of questions with possible additions to the PR that we could do before merging, if given a go-ahead.

Currently the cache is (as it was previously) only populated by the cache warmer. From my perspective, being able to save data into an empty cache would be a big boon, since that means the cache can warm itself up automatically.

* Would the client writing to the cache be acceptable to you?
* If yes, is the client writing to the cache considered a BC break, and as such require an option flag to activate?


magnusnordlander:
I have no idea what's up with Codeclimate. Old, deleted files are still there, and it's not installing the new dependencies from composer.json.

magnusnordlander:
Also, I just noticed that the Symfony Bundle extends the Warmer and Clearer, despite these not being `@api` marked, thus this PR is a BC break. As such, a third question arises:

* How willing/unwilling are you to release a new major version?

dborsatto:
Hey @magnusnordlander,
first of all, thanks for the contribution! 🎉 

Let's go through your questions:
- I like the idea of the SDK populating the cache during runtime, but it should definitely be an option and not the default behavior.
- Don't worry about CodeClimate. It's there for just pointing out obvious flaws in PRs, I can manually approve the PR there, or even simply ignore it.
- About the `@api` notation, I'm of the opinion that if a method is public, it's bound to be part of the public API of the SDK. Because of this, a breaking change to any public method would mean a new major version.

Now, I'm already planning 2 new major versions, one to be published soon-ish with a small BC break that's needed to add support for an upcoming Contentful functionality, and another (which doesn't have an ETA yet) which will include a quite big refactoring of the codebase. Because of this, as you can imagine, I'm not against having new major versions. The point is, it needs to make sense.

I have yet to take a look at the actual changes in the PR, but I just wanted to already give you some sort of feedback. To be honest, right now I'm working on something else, but I think I'll be able to get to this by the end of the week 🙂

dborsatto:
One thing I forgot, I'm definitely in favor of moving towards a PSR-6 (or PSR-16) cache implementation, rather than having a customized one. So even if doesn't happen right now (I'm not saying it won't, I'm saying _IF_ 🙂  ) rest assured that this is something in my todo list 👍

magnusnordlander:
Well, in that case, since this removes (now unnecessary) classes, this is a breaking change anyway. Unless of course we want to handle those removals the same way Symfony does and deprecate them first etc, but that seems like overkill. I'll let you be the judge of which breaks are acceptable and not.

Please take a look at the code when you have some time (we're not in that big a hurry), I'm more than happy to make changes. In the mean time, I'll work on populating the cache at runtime, and corresponding changes to the Symfony Bundle.

dborsatto:
Ok, some comments as I go through the PR:

Please don't remove `symfony/filesystem` from the dependencies. There are other packages that depend on this, and rely on this transitive dependency (even though it's admittedly not the best behavior). I'll get to this in the future, but in the meanwhile let's just leave it there.

I'm conflicted about how you defined the `symfony/cache` dependency. On the one hand, the _real_ dependency is on `psr/cache`, which is correctly locked to a version, which transitively forces `symfony/cache` to be locked to a version which works with `psr/cache 1.0`. This is a smart approach, but I'm not a fan of using `*` as stability modifier, so I'll have to think about it 😄 

---

Personally, I like PSR-16 more. I'm one of those people who thought that PSR-6 was overkill for most situations, and I was happy to see PSR-16 come to life. The previous PSR-6 PR which was rejected mainly because of how it was changing the behavior of `InstanceCache`, something which this PR is not doing.

With that said... It's probably best to have the internals of the SDK use PSR-6, whose added complexity is useful especially in the context of the `CacheWarmer`, where you can use `saveDeferred` and commit after the `foreach` is done.

One thing we can definitely do, however, is to recommend using `Symfony\Component\Cache\Adapter\SimpleCacheAdapter` so it can be effectively PSR-independent.

---

I'm not a fan of having 2 parameters in the `Client` constructor which handle caching. If we move towards this direction (which I'm definitely in favor of) I'd simply remove the `cacheDir` part and have a single parameter which expects either `null` (in which case a `NullCache` is created) or an object which extends `Psr\Cache\CacheItemPoolInterface`, whose configuration is delegated to the user. I'd prefer this to a duality of possible configurations, which could possibly cause confusion in users.

---

One of the reasons of moving to a PSR implementation of the cache would be the possibility of have the warmer be able to warm the cache with a configured `CacheInterface` implementation. Right now, we're using `FilesystemCache`, and your proposal doesn't really change that (it just replaced the previous implementation with `FilesystemAdapter`.

The problem in this case would be to find an elegant way of handling this through the CLI: it's easy to pass a directory path, but configuration parameters vary a lot for different adapters (think Redis, PDO, array, filesystem...). I'm thinking of possible ways of doing this, for instance making `CacheWarmer` and `CacheClearer` abstract, and then require implementations to have a method which returns a cache object, or maybe require some sort of factory path, or the path to a proxy object... I don't know, there are multiple ways we could handle this, if you have a good idea, feel free to post it! Of course, this solution should be something that can be easily handled by the Symfony integration.

---

Now that I think about it: considering the fact that I want to move the configuration of the cache out of the `Client`, we could probably also drop the dependency on `symfony/cache`. We could provide a PSR-6 implementation of `NullCache`, and use that as fallback, if the user doesn't provide one. We could move `symfony/cache` to suggest (and `require-dev` for testing purposes), so it should be a win-win.

---

This is *really* nitpicking (so feel free to ignore me on this one), but I don't like having static methods. I'm used to having `phpmd` telling me that static methods are bad -- but they state that for factory methods it's ok to use, and in this case we _could_ see `CacheKeyGenerator` as a factory, in a way. So if you can come up with anything else I'd be very happy, but as I said, this is nitpicking.

Still on the topic of generating keys, something I've been meaning to change is actually how the keys are formatted. Right now they're like this:
```
space.json
ct-contentTypeId1.json
ct-contentTypeId2.json
// ...
```
but for consistency (and arguably clarity) I wanted to change them to this:
```
space-spaceId.json
contentType-contentTypeId1.json
contentType-contentTypeId2.json
// ...
```

If you could change this directly in the PR, it would be great!

---

That's it! Sorry for the huge comment 😁  As I'm about to work on something that will require a BC break anyway, this is the perfect time to tackle this. If I said any nonsense or you disagree with me, feel free to voice your opinion 👍

magnusnordlander:
Thanks!

A couple of things:

* The cache warming is indeed an interesting problem. Getting a good integration in the Symfony bundle does not seem like a huge problem to me, I'm more worried about making it too difficult to use the library without a framework. I'll come up with something, but I'm guessing we may have to iterate this a couple of rounds.
* How do you feel about just using a couple of namespaced functions then for generating the cache keys? I dislike static methods as much as anyone, but there's really no point of requiring an instance of that class :)

I'll update the PR in response to your comments in the beginning of next week.

dborsatto:
I'm ok with moving methods to the `src/functions.php` file, but then again, it's the least of our concerns regarding this PR 🙂

src/functions.php
```diff
@@ -28,3 +28,16 @@ function format_date_for_json(\DateTimeImmutable $date)
...
+     return \sprintf('space-%s.json', $id);
+ }
+
+ /**
+  * @param $id
+  */
+ function content_type_cache_key($id)
+ {
+     return \sprintf('contentType-%s.json', $id);
+ }
```
dborsatto: I don't think that the cache keys should have the `.json` extension, because it's something specific to a cache implementation.

src/Delivery/Console/WarmUpCacheCommand.php
```diff
@@ -43,11 +43,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $cachePoolFactoryClass = $input->getArgument('cache-item-pool-factory-class');
+        $cacheItemPoolFactory = new $cachePoolFactoryClass();
+        if (!$cacheItemPoolFactory instanceof CacheItemPoolFactoryInterface) {
+            throw new \InvalidArgumentException("Cache item pool factory class must implement \Contentful\Console\CacheItemPoolFactoryInterface");
+        }
 
         $client = new Client($token, $spaceId);
-        $cache = new FilesystemAdapter($spaceId, 0, $cacheDir);
-        $warmer = new CacheWarmer($client, $cache);
+        $warmer = new CacheWarmer($client, $cacheItemPoolFactory->getCacheItemPool());
```
dborsatto: I would add a check here, to be sure that the result of `$cacheItemPoolFactory->getCacheItemPool()` actually is an instance of `CacheItemPoolInterface`.

composer.json
```diff
@@ -10,12 +10,14 @@
         "guzzlehttp/psr7": "~1.4",
         "psr/http-message": "~1.0",
         "psr/cache": "^1.0",
-        "symfony/cache": "*"
+        "symfony/filesystem": "~2.7|~3.0|~4.0",
+        "cache/void-adapter": "^1.0"
```
dborsatto: 👍

composer.json
```diff
@@ -10,12 +10,14 @@
...
     },
     "require-dev": {
         "php-vcr/phpunit-testlistener-vcr": "^2.0",
         "jakub-onderka/php-parallel-lint": "^0.9.2",
-        "jakub-onderka/php-console-highlighter": "^0.3.2"
+        "jakub-onderka/php-console-highlighter": "^0.3.2",
+        "cache/array-adapter": "^1.0"
     },
     "autoload": {
         "files": ["src/functions.php"],
```
dborsatto: Can you add a `suggest` block [like this](https://github.com/contentful/contentful-management.php/blob/master/composer.json#L36-L38) to suggest using `symfony/cache`?

src/Delivery/Client.php
```diff
@@ -114,13 +115,15 @@ public function __construct($token, $spaceId, $preview = false, $defaultLocale =
         $this->preview = $preview;
         $this->instanceCache = new InstanceCache();
         $this->defaultLocale = $defaultLocale;
+        $this->spaceId = $spaceId;
+
+        $this->cacheItemPool = null === $cache ? new VoidCachePool() : $cache;
```
dborsatto: You can skip the previous `$cache = $options['cache'];` and simply have
``` php
$this->cacheItemPool = $options['cache'] ?: new VoidCachePool();
```

src/Console/CacheItemPoolFactoryInterface.php
```diff
@@ -0,0 +1,15 @@
...
+
+interface CacheItemPoolFactoryInterface
+{
+    public function __construct();
+
+    /**
+     * @return CacheItemPoolInterface
+     */
+    public function getCacheItemPool();
+}
```
dborsatto: Even though I understand you reasoning for having this interface in this namespace, it would be a slight architectural problem, could you please move it to `Contentful\Delivery\Cache`? Thanks!

src/Console/CacheItemPoolFactoryInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Contentful\Console;
+
+use Psr\Cache\CacheItemPoolInterface;
+
+interface CacheItemPoolFactoryInterface
+{
+    public function __construct();
```
dborsatto: I think we should pass `$spaceId` as parameter here. Some cache implementations could use that for namespacing the cache items, so it's best to at least provide that option.

src/Delivery/Console/ClearCacheCommand.php
```diff
@@ -24,25 +24,23 @@ protected function configure()
...
-        $cacheDir = $input->getArgument('cache-dir');
+        $cachePoolFactoryClass = $input->getArgument('cache-item-pool-factory-class');
+        $cacheItemPoolFactory = new $cachePoolFactoryClass();
+        if (!$cacheItemPoolFactory instanceof CacheItemPoolFactoryInterface) {
+            throw new \InvalidArgumentException("Cache item pool factory class must implement \Contentful\Console\CacheItemPoolFactoryInterface");
+        }
 
-        $cache = new FilesystemAdapter($spaceId, 0, $cacheDir);
-        $clearer = new CacheClearer($cache);
+        $clearer = new CacheClearer($cacheItemPoolFactory->getCacheItemPool());
```
dborsatto: I would add a check here, to be sure that the result of `$cacheItemPoolFactory->getCacheItemPool()` actually is an instance of `CacheItemPoolInterface`.

src/Console/CacheItemPoolFactoryInterface.php
magnusnordlander: Noted, but I since I think it would be better to pass that in the `getCacheItemPool()` method in that case, I'll put it there for now :)

magnusnordlander:
There. All your comments have been adressed (though, as I said in a previous comment, I did a minor modification to one), and I've also made a PR for the Symfony bundle. 

Let the games begin!

dborsatto:
Ok, I'm going to do a final review of the while thing 😁

dborsatto:
Ok so, there is currently one problem.

`cache/void-adapter` and `cache/array-adapter` currently require PHP 5.6 or higher, whereas this SDK is currently supporting 5.5 or higher. In the next few days I'll be having some discussions internally about bumping the minimum required version of PHP to `7.0`, in the meanwhile I'm pausing this review. I'll get back to you as soon as we decide on what to do 🙂

magnusnordlander:
Cool. If it comes to it, it's not too difficult to write a void cache implementation, but since 5.5 has been EOL:ed for like a year and a half, it would be nice not to have to :)

dborsatto:
Hey @magnusnordlander,

sorry for taking so long 🙂 

After some internal discussions, I've decided that this PR can be merged as-is. Though we haven't decided whether to drop 5.5 yet (hopefully soon), I don't want to have this extremely nice PR become stale, so I'd like to merge it and then I'll make a couple changes for cleanup, like removing 5.5 in Travis. Adding it back shouldn't be a huge problem, I'd have to drop `cache/*-adapter` from the dependencies and build an internal one, but it's no big deal.

There's one thing I have to ask you: unfortunately I had to make some changes that made this branch out of date, so I have to ask you a final rebase on top of the latest changes of `master`. There are 4 conflicting files, but 3 have been removed (so just mark them as such) and in `composer.json`I removed the dependencies on the `jakub-onderka` packages, so make sure to reflect that change in your branch.

Let's get this merged! 😎

magnusnordlander:
No problem, I'll get it rebased right now :)

magnusnordlander:
There. Rebased :)

However, since this is a BC breaking change, the branch alias in composer.json should be updated to 3.0.0-dev. Do you want me to do it, or do you want to do it after merging?

PR #191, opened 2018-02-28
Add support for environments

dborsatto:


CHANGELOG.md
```diff
@@ -7,21 +7,32 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
 
 ### Changed
 
 * The cache system has been rewritten to be made PSR-6 compatible (thanks @magnusnordlander). **[BREAKING]**
-* `DynamicEntry`, `Asset`, `ContentType`, `ContentTypeField`, `Space` and `Locale` classes have been moved to a different namespace. Please check the [the upgrade guide](UPGRADE-3.0.md) for more details. **[BREAKING]**
+* `DynamicEntry`, `Asset`, `ContentType`, `ContentTypeField`, `Space`, and `Locale`, `DeletedAsset`, 'DeletedEntry', and `DeletedContentType` classes have been moved to a different namespace. Please check the [the upgrade guide](UPGRADE-3.0.md) for more details. **[BREAKING]**
+* All parts of the SDK that were not in the `Contentful\Delivery` namespace have been moved to a separate package called [contentful-core.php](https://github.com/contentful/contentful-core.php).
+* The option in the client constructor for specifying a custom URI is not called `baseUri` instead of `uriOverride`, to be more consistent with the one used in Guzzle. **[BREAKING]**
```
dlitvakb: It should say `it's *now* called` instead of `it's *not* called`.

README.md
```diff
@@ -1,103 +1,99 @@
...
 
-```php
-$client = new \Contentful\Delivery\Client('access-token', 'space-id', true);
+``` php
+$client = new \Contentful\Delivery\Client('access-token', 'space-id', 'environment-id', true);
 ```
 
-Apart from the configuration option, you can use the SDK without modifications with one exception: you need to obtain a preview access token, which you can get in the "API" tab of the Contentful app. In preview mode, data can be invalid, because no validation is performed on unpublished entries. Your app needs to deal with that. Be aware that the access token is read-write and should in no case be shipped with a production app.
+Apart from the configuration option, you can use the SDK without modifications with one exception: you need to obtain a preview access token, which you can get in the "API" tab of the Contentful web  app. In preview mode, data can be invalid, because no validation is performed on unpublished entries, so tour app needs to take that into account.
```
dlitvakb: `so your` instead of `so tour`

src/Client.php
```diff
@@ -381,6 +426,43 @@ private function extractSpaceId(array $data)
...
+        if (isset($data['items'][0]['sys']['environment'])) {
+            return $data['items'][0]['sys']['environment']['sys']['id'];
+        }
+
+        // Empty array resource
+        if (isset($data['items']) && !$data['items']) {
+            return $this->environmentId;
+        }
+
+        return '[blank]';
```
dlitvakb: Why `'[blank]'` and not `'master'`?

src/Mapper/Space.php
```diff
@@ -28,18 +27,6 @@ public function map($resource, array $data)
         return $this->hydrate($resource ?: ResourceClass::class, [
             'sys' => $this->buildSystemProperties($data['sys']),
             'name' => $data['name'],
-            'locales' => \array_map([$this, 'buildLocale'], $data['locales']),
```
dlitvakb: Space still has locales

src/Resource/Environment.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * @param string $code Code of the locale to fetch the object for
+     *
+     * @throws \InvalidArgumentException When no locale with the given code exists
+     *
+     * @return Locale
+     */
+    public function getLocale($code)
```
dlitvakb: I personally don't think that this should be here, as it is a new endpoint.

src/Resource/Space.php
```diff
@@ -19,11 +19,6 @@ class Space extends BaseResource
      */
     protected $name;
 
-    /**
-     * @var Locale[]
-     */
-    protected $locales = [];
```
dlitvakb: Locales are still part of the space

src/SystemProperties.php
```diff
@@ -186,6 +201,15 @@ public function jsonSerialize()
                 ],
             ];
         }
+        if (null !== $this->environment) {
```
dlitvakb: I'd avoid adding a workaround for environments unless strictly necessary

src/Client.php
dborsatto: I'm debating this.

Using `master` would be helpful for backwards compatibility, because a pre-environments JSON string would be correctly recognized, but on the other hand if you reach this point you have done something weird.

I think I'll change it to `master` for now, but I'll probably deprecate this sometimes in the future 👍

src/Mapper/Space.php
dborsatto: Only for backwards compatibility purposes in the API, the handling of locales should happen at the environment level, from now on...

src/Resource/Environment.php
dborsatto: It's a special case... You don't have `/spaces/xxx/environments/yyy/locales/zzz`, you only have the collection endpoint.

To me it makes sense to keep it here, as it's consistent with the previous SDK behavior, and if you want to get locales you can simply do:

``` php
$locales = $client->getEnvironment()->getLocales();
```

It's functionally the same, with the advantage that with this approach, I can much more easily handle caching, because of inherent limitations of working with a collection endpoint rather than using a regular ID for querying.

src/Resource/Space.php
dborsatto: See me comment above

src/SystemProperties.php
dborsatto: Not a workaround, that's how everything is handled in `SystemProperties`

PR #195, opened 2018-03-05
Replace cache with repository

dborsatto:


src/Client.php
```diff
@@ -183,9 +176,14 @@ protected function getApiContentType()
      */
     public function getAsset($assetId, $locale = null)
     {
-        $locale = null === $locale ? $this->defaultLocale : $locale;
+        $locale = $locale ?: $this->defaultLocale;
 
-        return $this->requestAndBuild('GET', 'assets/'.$assetId, [
+        $instanceId = $assetId.'-'.($locale ?: '*');
```
loudmouth: Should this be abstracted into a `getCacheKey` method?

dborsatto: Possibly, yeah 🤔

dborsatto: But then I would need to supply the locales info to the `getCacheKey` method...

src/Client.php
```diff
@@ -92,49 +90,44 @@ class Client extends BaseClient
...
      */
     public function __construct($token, $spaceId, $preview = false, $defaultLocale = null, array $options = [])
     {
-        $baseUri = $preview ? 'https://preview.contentful.com/' : 'https://cdn.contentful.com/';
-
         $options = \array_replace([
             'guzzle' => null,
             'logger' => null,
-            'uriOverride' => null,
+            'baseUri' => null,
```
dlitvakb: What's the reasoning for this change?

I agree that this is a better name, but this will add one more thing that users will have to have in mind if they want to upgrade to this new version.

Therefore, please create a guide for upgrading, and include this change in the `Changes` section of the changelog.

src/Client.php
```diff
@@ -92,49 +90,44 @@ class Client extends BaseClient
...
         ], $options);
 
-        $guzzle = $options['guzzle'];
-        $logger = $options['logger'];
-        $uriOverride = $options['uriOverride'];
-        $this->autoWarmup = $options['autoWarmup'];
-
-        if (null !== $uriOverride) {
-            $baseUri = $uriOverride;
+        $baseUri = $preview ? self::URI_PREVIEW : self::URI_DELIVERY;
```
dlitvakb: This should be extracted to a method.

src/Client.php
```diff
@@ -92,49 +90,44 @@ class Client extends BaseClient
...
 
         $this->preview = $preview;
-        $this->instanceCache = new InstanceCache();
         $this->defaultLocale = $defaultLocale;
         $this->spaceId = $spaceId;
 
-        $this->cacheItemPool = $options['cache'] ?: new VoidCachePool();
-
-        if (!$this->cacheItemPool instanceof CacheItemPoolInterface) {
+        $cacheItemPool = $options['cache'] ?: new VoidCachePool();
```
dlitvakb: Do you do this as a way to turn off caching or is it just to provide a default?

If a default is desired, it should probably not be a VoidCache, but rather an in memory cache.

If having a null cache is desired, then this is OK.

Please clarify.

src/Client.php
```diff
@@ -183,9 +176,14 @@ protected function getApiContentType()
...
-        $locale = null === $locale ? $this->defaultLocale : $locale;
+        $locale = $locale ?: $this->defaultLocale;
 
-        return $this->requestAndBuild('GET', 'assets/'.$assetId, [
+        $instanceId = $assetId.'-'.($locale ?: '*');
+        if ($this->instanceRepository->has('Asset', $instanceId)) {
+            return $this->instanceRepository->get('Asset', $instanceId);
+        }
+
+        return $this->requestAndBuild('/spaces/'.$this->spaceId.'/assets/'.$assetId, [
```
dlitvakb: The `'/spaces'.$this->spaceID` could be extracted to the `requestAndBuild` method, as there is no endpoint in the CDA that doesn't require that path.

src/InstanceRepository.php
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * @param string $type
+     * @param string $resourceId
+     *
+     * @return bool
+     */
+    public function has($type, $resourceId)
+    {
+        $this->warmUp($type, $resourceId);
```
dlitvakb: What happens if this is run on a collection of entries? Will it try to request each individual entry?

src/Client.php
dborsatto: There's already a `UPGRADE-3.0.md` file, I have a bunch of other stuff I have to add to it. It's not in this PR, but it's going to be there 🙂

src/Client.php
dborsatto: It's a default for not storing anything in cache. It adheres to the interface that `InstanceRepository` requires, except it doesn't retain anything in memory.

src/Client.php
dborsatto: No, I actually want to have the full URL. Previously it would set the `/spaces/.$this->spaceId` as part of the HTTP client base URI options, but I changed that behavior because I'd rather have consistency than save a few characters.

src/Client.php
dborsatto: I can think about it, but it'd be probably overkill...

dlitvakb: Not this particular line, I mean the whole snippet of `baseUri` mangling

PR #202, opened 2018-04-10
Refactor __call in Entry resource

dborsatto:
Refactors all the logic in the `__call` method to smaller, more maintainable, and less troublesome methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 * The SDK now keeps a registry of all resources that are currently managed, called `Contentful\Delivery\InstanceRepository`. This class also wraps the PSR-6 cache pool.
 * `Client::reviveJson()` is now called `Client::parseJson()` to better reflect its meaning.
 * The Sync API currently only works with the `master` environment. When trying to perform sync-related operations on a client which is configured with any other environment, a `\RuntimeException` will be thrown.
+* Previously, entry objects would try to use the locale fallback chain even in situations where this might lead to results that differ from those that would be normally returned from the Delivery (or Preview) API. For this reason, when requesting an entry using any other locale than `locale=*`, the locale fallback chain will not be used. The Sync API defaults to using all locales, so it's not affected. **[BREAKING]**
 
 ### Removed
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -115,7 +115,6 @@ class Client extends BaseClient
             'guzzle' => null,
             'logger' => null,
             'baseUri' => null,
-            'cacheDir' => null,
             'cache' => null,
             'autoWarmup' => false,
         ], $options);
--- a/src/Mapper/BaseMapper.php
+++ b/src/Mapper/BaseMapper.php
@@ -72,13 +72,7 @@ abstract class BaseMapper implements MapperInterface
 
         if ($target instanceof LocalizedResource) {
             $locales = $this->client->getEnvironment()->getLocales();
-            $target->setLocales($locales);
-
-            /** @var SystemProperties $sys */
-            $sys = $target->getSystemProperties();
-            if ($locale = $sys->getLocale()) {
-                $target->setLocale($locale);
-            }
+            $target->initLocales($locales);
         }
 
         return $target;
--- a/src/Resource/ContentType.php
+++ b/src/Resource/ContentType.php
@@ -82,16 +82,25 @@ class ContentType extends BaseResource
      * If the field does not exist, null is returned.
      *
      * @param string $fieldId
+     * @param bool   $tryCaseInsensitive
      *
      * @return Field|null
      */
-    public function getField($fieldId)
+    public function getField($fieldId, $tryCaseInsensitive = false)
     {
-        if (!isset($this->fields[$fieldId])) {
-            return null;
+        if (isset($this->fields[$fieldId])) {
+            return $this->fields[$fieldId];
         }
 
-        return $this->fields[$fieldId];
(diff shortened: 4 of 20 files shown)


src/Client.php
```diff
@@ -115,7 +115,6 @@ public function __construct($token, $spaceId, $environmentId = 'master', $previe
             'guzzle' => null,
             'logger' => null,
             'baseUri' => null,
-            'cacheDir' => null,
```
dborsatto: This was a leftover from a previous PR

src/Mapper/BaseMapper.php
```diff
@@ -72,13 +72,7 @@ protected function hydrate($target, array $data)
 
         if ($target instanceof LocalizedResource) {
             $locales = $this->client->getEnvironment()->getLocales();
-            $target->setLocales($locales);
-
-            /** @var SystemProperties $sys */
-            $sys = $target->getSystemProperties();
-            if ($locale = $sys->getLocale()) {
-                $target->setLocale($locale);
-            }
```
dborsatto: This is now handled in `initLocales`

src/Resource/Entry.php
```diff
@@ -23,11 +23,6 @@ class Entry extends LocalizedResource
      */
     protected $fields;
 
-    /**
-     * @var array
-     */
-    protected $resolvedLinks = [];
```
dborsatto: Previously entries would keep a locale cache of resolved links, but now the SDK uses a global entry registry so there's no need to keep doing this locally

tests/E2E/AssetTest.php
```diff
@@ -97,10 +97,6 @@ public function testIncludedAssetLocale()
         $this->assertSame('Ben Chang', $asset->getTitle());
         $this->assertSame('I AM A SPANISH GENIUS!', $asset->getDescription());
 
-        // Tests that the fallback chain is working correctly.
-        // Tested chain is es -> it -> en-US
-        $this->assertNull($asset->getFile('it'));
-        $this->assertNull($asset->getFile('en-US'));
```
dborsatto: These assertions actually relied on a faulty behavior.

tests/TestCase.php
```diff
@@ -247,8 +254,10 @@ public static function withSys($id, $data = [])
 
 class MockField extends \Contentful\Delivery\Resource\ContentType\Field
 {
-    public function __construct(array $data)
+    public function __construct($id, $name, $type, array $data = [])
```
dborsatto: I've changed the constructor of `MockField` for ease of use, because ID, name and type are always set either way.

src/Resource/Entry.php
```diff
@@ -108,215 +103,196 @@ public function __call($name, $arguments)
...
+        // we could be returning this in the try/catch without creating a variable,
+        // but for some reason the coverage reporter will detect a non-covered
+        // closing bracket at the end of the method...
         return $result;
     }
 
     /**
-     * Resolves a Link into an Entry or Asset.
-     * Resolved links are cached local to the object.
+     * Returns a the value of a field using the given locale.
```
dborsatto: From this point forward, Git is trying to match old methods and new ones.

In reality though, they don't match at all, so disregard the line matching attempts being made here.

`resolveLinkWithCache`, `getFieldConfigForName`, `mapValues`, `mapIdValues` are not needed because of how field getters work now.

`formatSimpleValueForJson` and `formatValueForJson` were simply useless to being with.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 * The SDK now keeps a registry of all resources that are currently managed, called `Contentful\Delivery\InstanceRepository`. This class also wraps the PSR-6 cache pool.
 * `Client::reviveJson()` is now called `Client::parseJson()` to better reflect its meaning.
 * The Sync API currently only works with the `master` environment. When trying to perform sync-related operations on a client which is configured with any other environment, a `\RuntimeException` will be thrown.
+* Previously, entry objects would try to use the locale fallback chain even in situations where this might lead to results that differ from those that would be normally returned from the Delivery (or Preview) API. For this reason, when requesting an entry using any other locale than `locale=*`, the locale fallback chain will not be used. **[BREAKING]**
```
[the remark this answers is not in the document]
dborsatto: I thought it was obvious but you're right, better to make it explicit 🙂

src/Resource/ContentType.php
```diff
@@ -82,16 +82,25 @@ public function getDescription()
...
-    public function getField($fieldId)
+    public function getField($fieldId, $tryCaseInsensitive = false)
     {
-        if (!isset($this->fields[$fieldId])) {
-            return null;
+        if (isset($this->fields[$fieldId])) {
+            return $this->fields[$fieldId];
+        }
+
+        if ($tryCaseInsensitive) {
```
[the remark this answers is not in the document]
dborsatto: To be honest I'd rather not, this is a small "hack" that I used to be able to simplify management in the entry class, in order to keep compatibility with some older behavior (which I don't like too much). Because of this, let's keep it "hidden" 😅

src/Resource/Entry.php
```diff
@@ -108,215 +103,196 @@ public function __call($name, $arguments)
...
-            if ($getId) {
-                return \array_map([$this, 'mapIdValues'], $result);
-            }
-
-            return \array_filter(\array_map([$this, 'mapValues'], $result), function ($value) {
-                return !$value instanceof \Exception;
-            });
-        }
-
+        // This is a ugly hack to trick the coverage reporter;
```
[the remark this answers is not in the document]
dborsatto: You're right 👍

PR #122, opened 2017-05-31
Don't attempt to deal with linked fields that are present but empty objects

Jalle19:
This only happens when using web hooks

Fixes #103, closes #102 because it won't be required for #103

PR #205, opened 2018-04-11
Prevent empty fields from being built in entries

dborsatto:
Replaces #122

### DIFF
--- a/tests/Integration/NullValueTest.php
+++ b/tests/Integration/NullValueTest.php
@@ -29,4 +29,21 @@ class NullValueTest extends TestCase
 
         $this->assertNull($entry->getLogo());
     }
+
+    /**
+     * @see https://github.com/contentful/contentful.php/issues/103
+     * @see https://github.com/contentful/contentful.php/pull/122
+     */
+    public function testFieldIsPresentButEmpty()
+    {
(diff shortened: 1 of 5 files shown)


tests/Integration/NullValueTest.php
```diff
@@ -29,4 +29,21 @@ public function testNullAsValueForLink()
...
+    {
+        $client = new Client('irrelevant', 'rue07lqzt1co');
+
+        $client->parseJson($this->getFixtureContent('space.json'));
+        $client->parseJson($this->getFixtureContent('environment.json'));
+        $client->parseJson($this->getFixtureContent('content_type.json'));
+        $entry = $client->parseJson($this->getFixtureContent('entry_field_without_values.json'));
+
+        $this->assertNull($entry->get('logo'));
+        $this->assertSame([], $entry->get('phone'));
```
dlitvakb: Why is phone an array here?

dlitvakb: I just saw the Content Type definition, alles gut!

dborsatto: Because in the content type it's defined as an array. The meaningfulness of the whole thing is not important, I just needed one field defined as array, and another defined as anything else 😉

PR #206, opened 2018-04-16
Optimizations to system properties

dborsatto:


src/SystemProperties.php
```diff
@@ -122,9 +161,25 @@ public function getSpace()
...
         return $this->contentType;
     }
 
+    /**
+     * @return Environment|null
+     */
+    public function getEnvironment()
+    {
+        if ($this->environment instanceof Link) {
+            $this->environment = $this->client->resolveLink($this->environment);
```
dlitvakb: How are you solving this? You require both space ID and environment ID. Not just Environment ID.

Similar question for Content Types, for which you need space ID, environment ID and content type ID.

dborsatto: The client is configured with space ID and environment ID, so it's not really resolving this link, because the client can't actually query anything else but that space or that environment.

I could have written `$this->environment = $this->client->getEnvironment();` (and the same for space), but I chose to do it like so simply for having some consistency in terms of how links appear to be resolved 🙂

dlitvakb: So, for each new environment I'd want to work with, I'd have to create a new client instance or can I still use the same (without altering the client configuration)?

dborsatto: Yes, the CDA SDK is built with the assumption that to one client corresponds 1 space/1 environment. It simplified a lot of stuff architecturally speaking!

dlitvakb: Cool :D

PR #209, opened 2018-04-24
Run static analysis using Phpstan

Jalle19:
Figured it can't hurt. Going above level 2 requires some major type-hint changes so I didn't bother. Level 1 catches the biggest issues anyway so it should be fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,14 @@ script:
     fi
 
   - |
+    # Static analysis (build lead only since phpstan requires PHP >= 7.0)
+    if [ $BUILD_LEAD == 1 ]
+    then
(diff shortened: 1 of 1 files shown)


dborsatto:
I'm already running PHPStan locally (without using that as a dependency), and unfortunately because of the generic nature of certain parts of the SDK, going above level 2 causes major type inference problems. PHPStan does its best, but they're definitely false positives (that's where actual tests come into play)...

I'm waiting for the build status to see if it works 😁 if it does, then it's definitely a 👍

Jalle19:
Since you support PHP 5.6 I couldn't pull it from Composer, but the current approach should work

dborsatto:
Yes, it builds 🎉 
Do you want to change something or can I merge? 🙂

Jalle19:
@dborsatto feel free to merge

dborsatto:
Thanks @Jalle19 🙂

PR #220, opened 2018-06-06
Change entry mapper to reuse previous fields from entry object

dborsatto:


src/Mapper/Entry.php
```diff
@@ -44,13 +45,23 @@ public function map($resource, array $data)
      * @param ResourceContentType $contentType
      * @param array               $fields
      * @param string|null         $locale
+     * @param ResourceClass|null  $previous
```
dlitvakb: Why is a class called `Class`?

src/Mapper/Entry.php
```diff
@@ -80,16 +94,65 @@ private function buildFields(ResourceContentType $contentType, array $fields, $l
     }
 
     /**
-     * @param ResourceContentTypeField $fieldConfig
-     * @param array                    $fieldData
+     * If a previous entry is give, this methods will combine all values from the $fields
```
dlitvakb: If a previous entry is `given`

src/Mapper/Entry.php
```diff
@@ -80,16 +94,65 @@ private function buildFields(ResourceContentType $contentType, array $fields, $l
...
+        // As we can't know for certain which fields will be present (some might not be available
+        // in the current API response, some might be excluded anyway because they're blank...)
+        // we just use the field definitions from the content type and progressively check which fields
+        // are actually present.
+        foreach ($contentType->getFields() as $field) {
+            $fieldId = $field->getId();
+            $fieldLocales = $field->isLocalized() ? $locales : [$defaultLocale];
+
+            foreach ($fieldLocales as $locale) {
+                if (!isset($fields[$fieldId][$locale]) && $previous->has($fieldId)) {
```
dlitvakb: Is it possible to ensure with a revision check that the object you're updating has not been changed?

PR #227, opened 2018-07-26
Update to notify upcoming PHP 7 requirements

dborsatto:


michaelorr:
@dborsatto Can you elaborate on the plans surrounding this move? Specifically, when the jump from 3.x to 4.x is expected and what maintenance 3.x will receive once 4.x is live?

PR #239, opened 2018-09-20
Add specific sys classes for every resource

dborsatto:


tests/E2E/EntryTest.php
```diff
@@ -155,8 +155,8 @@ public function testAssetsResolvedFromIncludes()
         $this->assertSame('nyancat', $image->getId());
 
         // There should be 3 and only 3 requests:
```
mariobodemann: 4 and only 4?

dborsatto: Oh yeah, I forgot to update the comment 😁

PR #245, opened 2018-10-19
Do not call API for included assets when using * for locale

Spea:
I wanted to upgrade from `2.4.2` to `3.4.0` but then I realized that we now have one call for every asset in the fetched entries. After some debugging I found out that when using the `?locale=*` parameter, the cache key always uses the `__ALL__` suffix, but when requesting the asset the current locale is used (in this case `en-US`) which does not exist in the resource pool.

I then decided to adjust the ResourcePool so it also checks for the `__ALL__` key whenever the locale-specific one was not found. I'm not sure if this is the correct/perfect way, but every other solutions seemed more of a hack than this one.

### DIFF
--- a/src/ResourcePool.php
+++ b/src/ResourcePool.php
@@ -131,6 +131,13 @@ class ResourcePool extends BaseResourcePool
         $key = $this->generateKey($type, $id, $options);
         $this->warmUp($key, $type);
 
+        if (isset($this->resources[$key])) {
+            return true;
+        }
+
+        unset($options['locale']);
+        $key = $this->generateKey($type, $id, $options);
+
         return isset($this->resources[$key]);
     }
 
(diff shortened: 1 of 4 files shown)


dborsatto:
Hey @Spea,

First of all, thanks for the contribution! Bear in mind, currently the master branch is receiving a lot of breaking changes and a new major release will happen within a few weeks, so whatever goes into master right now won't end up in the v3 major version.

About this specific change, I think this shouldn't happen. Entries and assets when with a locale and when without one are fundamentally different, and because of how fallback locales are handles, mixing the two things in the resource pool might lead to unpredictable results.

What is your use case, exactly?

Spea:
I also added a test which reflects my use case. 

I'm requesting entries with `locale=*` and when I do a `$entry->getImage()` a new call to the API is triggered even though the asset was already in the `includes` property from the entries-call. We have one entry with ~20 assets, and doing 20 extra calls makes the whole chain pretty slow.

dborsatto:
This sounds like a bug, but not where you're trying to fix with this PR. Assets in the include should have the same locale as the actual entries, so the locales should match. I'll take a look at this!

PR #246, opened 2018-10-19
Fix resource pool handling of linked resources from entries

dborsatto:
Closes #245


dborsatto:
@Spea this is the actual fix for your issue, the link resolving in the `Entry` class was the real culprit 🙂

Spea:
Amazing, thank you very much for the very quick fix :)

PR #273, opened 2019-12-02
Php 115 fix infinite loop

pgrigoruta:
When a rich text entry contained a reference to itself, the parser entered an infinite loop trying to expand the entry. Inspecting the Python implementation, I took the same approach of limiting the depth to 20.


pgrigoruta:
@matthew-contentful , please re-check. The linter is now fixed, I have also added test cases to get back to 100% coverage.

PR #276, opened 2019-12-10
PHP-52 - add back code climate

pgrigoruta:


pgrigoruta:
No functional changes here, adding back code climate

PR #284, opened 2020-03-03
Feature/php74 support

pgrigoruta:


pgrigoruta:
Added support for PHP 7.4, dropped support for PHP 7.0 and 7.1

Issue #294, opened 2020-09-28
BaseClient::messages memory exhaustion


brettmc:
I am using the SDK (v6.0.3) in a long-running CLI process, to periodically query contentful and process the response.

In tracking down what we originally thought was a memory leak, whereby our PHP process would eventually exhaust memory and die, I found that in fact the culprit was `Api\BaseClient::messages` storing all of the http requests and responses in an ever-growing collection (which can grow really fast if you're pulling down non-trivial amounts of data).

I can think of a couple of ways this could be avoided:
* provide a `ClientOptions::withoutMessageLogging()` method which controls whether messages are stored (maintaining the storing of messages as the default behaviour)
* adding an `Api\BaseClient::clearMessages()` method to allow clearing out on-demand

I'd be happy to submit a PR if I can get some guidance as to whether either of these options seem reasonable?

pgrigoruta:
Hi @brettmc , thanks for finding this! I think both approaches are good and serve different use cases, so best is to implement them both. If you have a PR (full or partial), please submit it and we can finish the implementation/add test coverage if needed.

brettmc:
step 1, I have submitted a PR against contentful-core.php (https://github.com/contentful/contentful-core.php/pull/43)
If that's all cool, then there will be a small follow-on patch in this repo

pgrigoruta:
Thanks @brettmc , I just merged this branch

PR #295, opened 2020-10-12
adding ClientOptions::withoutMessageLogging

brettmc:
As a follow-up patch to https://github.com/contentful/contentful.php/issues/294 this patch exposes the ability to disable the logging
of messages via ClientOptions. The default behaviour remains to continue to log messages.
Bumped the minimum required version of contentful/core to ^3.1 in anticipation that the next release will indeed be v3.1.
In passing, update the php-vcr version requirement to the latest version, as the branch it was trying to use no longer exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ $options = \Contentful\Delivery\ClientOptions::create()
     ->withLogger(Psr\Log\LoggerInterface $logger)
     ->withCache(Psr\Cache\CacheItemPoolInterface $cache, bool $autoWarmup = false, bool $cacheContent = false)
     ->withHttpClient(GuzzleHttp\Client $client)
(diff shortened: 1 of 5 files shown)


pgrigoruta:
Thanks you for your contribution @brettmc !

brettmc:
Hi @pgrigoruta - contentful/core lib definitely *does* have new changes that this patch relies on, namely https://github.com/contentful/contentful-core.php/pull/43. The patch won't work without those changes (I think it will silently do nothing) which is why I bumped the dependency up.
As a minimum, this change would require a new released version of contentful/core (composer may pick up the correct version anyway, but bumping the dependency to make it explicit felt like the "right way", according to semantic versioning).

Issue #269, opened 2019-08-22
[Question] Is the getEntries() request cached?


mmenozzi:
The `$client->getEntries()` is as follows:

https://github.com/contentful/contentful.php/blob/master/src/Client.php#L365-L380

It calls `$this->request()` and not `$this->requestWithCache`. So it seems is not cached. Is that true? Am I missing something?

mmenozzi:
@realityking @dborsatto do you have any advice on this?

dborsatto:
@mmenozzi The SDK can not really predict what a call to a collection endpoint might return: if (for instance) a user has added another entry using the Contentful web application, having the SDK use some sort of cache layer will inevitably lead to a stale result set.

For this reason, only calls to `getEntry`, `getAsset`, and `getContentType` will go through a cache layer, as the SDK is certain of what resource (entry, asset or content type) the user is requesting.

By the way, Rouven stopped working on the SDK back in 2017, and I was its maintainer until last February. Nowadays I'm not really sure who's keeping it alive, but I thought I'd help you with your question anyway.

Cheers,

Davide

mmenozzi:
Thanks Davide!
Anyway I do not agree. Even for the `getEntry` method there might be the same problem. If, using Contentful web application, I change a field value on a content already cached I will get the old value if that content is fetched with `getEntry` in my application.

And this is totally fine for me. It's my responsibility to invalidate that cache in some way and get new content.

So I could do the same for the `getEntries` method.

dborsatto:
The cache used by `getEntry` is by default not persisted across requests. This means that for every request the SDK will start from scratch, so there's no chance of getting a stale result.

The cache _can_ be made persistent (leading to good speed bumps for certain use cases) but it's an opt-in process through the `$options` object in the client constructor.

PR #306, opened 2022-01-03
Implement Client::getEntries() query cache layer

mmenozzi:
See also #269.

This PR is a first WIP to add caching also to `Client::getEntries()`. It adds a `QueryPoolInterface` and its `Standard` implementation which behaves like the `Extended` `ResourcePool`: with default configuration entries are not cached across different client instantiations (different requests). Similarly to the `cacheContent` option of the resource pool cache, is it also possible to set a lifetime which makes entries to be cached also across different client instantiations.

Before going on I'd like to know what the maintainers (@Sebb767) think about this.


Sebb767:
Hi @mmenozzi ,

first, thanks a lot for contributing! I agree with you that optionally caching `getEntries()` makes sense.

I haven't had time to fully test the PR yet, but it's looks good so far. One point though - do you think it makes sense to allow the client to have separate caching pools for entries and queries? In my opinion, simply adding a fourth parameter `enableQueryCaching` to `withCache()` [here](https://github.com/mmenozzi/contentful.php/blob/9abec5ce5f0143505a60fcac030dd64d90f52b97/src/ClientOptions.php#L141)¹ would be simpler, since then you would not need to supply the `CacheItemPoolInterface` twice. This would, however, prevent one from using different cache types for queries and entries - I'm not sure how common this would be, though.


 ¹ There could also be method `enableQueryCaching(int $queryCacheLifetime = 0)` instead or in addition to simplify the interface, but it's the same idea.

mmenozzi:
Hi @Sebb767,
I thought about implementing query caching in the current implementation of the `ResourcePoolInterface`, in the end a `ResourceArray` implements the `ResourceInterface`. But I then decided to do it with a separate `QueryPoolInterface` for the following reasons:

- Caching single resource is different than caching multiple resources gathered with a query. Think for example about the `ResourcePoolInterface::save()` method. How you'd implement it to save a `ResourceArray`? You need the Query which has been used to get those resources and the current signature of the save method does not allow to pass such input.
- For the reason above (and maybe for other reasons) a change to the `ResourcePoolInterface` would be needed that would led us to a breaking change.
- Having different pool for query caching with potentially different cache backend could be helpful. Again queries and their results are different from single resources. Potentially they are accessed more frequently and for sure they must store more data per single entry. So a different cache backend could be helpful.

Sebb767:
Hi @mmenozzi ,

this sounds reasonable. I'll include your changes in the next release.
Thanks a lot for contributing!

PR #307, opened 2022-01-12
Update composer.json

Kyoss79:
update psr/cache dependency

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "php": "^7.2|^8.0",
         "contentful/core": "^3.0",
         "contentful/rich-text": "^3.1",
-        "psr/cache": "^1.0",
+        "psr/cache": "^1.0|^3.x",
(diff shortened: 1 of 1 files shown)


Kyoss79:
Heyho, sorry I did this in your repository. ;-) 

Made a fork and tried it there.

The underlying issue was, that with a new version of laravel, contentful/laravel cannot be installed anymore without forcing composer to.

>  - contentful/contentful[6.1.0, ..., 6.2.0] require psr/cache ^1.0 -> found psr/cache[1.0.0, 1.0.1] but the package is fixed to 3.0.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.

Sebb767:
Thanks for contributing! I manually edited the composer.json, since I wanted to do some testing, but I'll include this change in the next release :)

PR #309, opened 2022-02-20
Replace cache/void-adapter with symfony/cache

kwivix:
Replace cache/void-adapter with symfony/cache because the void-adapter package is not updated any more. From the php cache main repository:

> Back in 2016, this was the first library supporting PSR-6. We created Symfony bundles and made many great libraries in the PHP-cache organization. This was a few years ago and the library is not activly maintained. Starting a new project one should consider using the more performant and activly supported [Symfony Cache](https://symfony.com/doc/current/components/cache.html).
> https://github.com/php-cache/cache#php-cache

See also https://github.com/contentful/contentful-core.php/pull/53

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,16 +7,16 @@
         "php": "^7.2|^8.0",
         "contentful/core": "^3.0",
         "contentful/rich-text": "^3.1",
-        "psr/cache": "^1.0|^3.0",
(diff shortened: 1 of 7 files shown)


Sebb767:
I'm on it; sorry for the delay :) If nothing major breaks (and it looks like it currently) I'll finish the releases by Thursday.

PR #316, opened 2022-09-13
Prevent invalid links from rendering an entry inaccessible

bullenb:
When a link associated to an entry cannot be resolved, the attempt to resolve the parent entity fails. This can even cause an attempt to fetch a collection of entries to fail.

This change assumes it is acceptable to produce a null value for an invalid link rather than throw an exception.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @contentful/team-tundra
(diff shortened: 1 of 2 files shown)


LucaGallinari:
Hi @Sebb767 we are interested in this PR to be merged too. If you can find some time to review it, it would be really appreciated 😁

adithya-flaconi:
Any luck on this PR? Would help us out!

Sebb767:
Merging this PR as-is would change the behavior of the library significantly; therefore we can't easily merge it. We're currently working on a way to allow rendering entries with invalid links via a configuration flag in order to fix this issue.

adithya-flaconi:
@Sebb767 Is that planned / already fixed in the main branch? thanks!

PR #317, opened 2022-09-16
PHP 8.1 Compatibility Fix - Add appropriate return types

dbebawy-modus:
I've added the appropriate return types to resolve the PHP Deprecation Notices in PHP 8.1.

### DIFF
--- a/src/Resource/Entry.php
+++ b/src/Resource/Entry.php
@@ -126,7 +126,7 @@ class Entry extends LocalizedResource implements EntryInterface, \ArrayAccess
     /**
      * {@inheritdoc}
      */
-    public function offsetGet($name)
+    public function offsetGet($name): mixed
     {
         return $this->get($name);
     }
@@ -134,7 +134,7 @@ class Entry extends LocalizedResource implements EntryInterface, \ArrayAccess
     /**
      * {@inheritdoc}
      */
-    public function offsetExists($name)
+    public function offsetExists($name): bool
     {
         return $this->has($name);
     }
@@ -142,7 +142,7 @@ class Entry extends LocalizedResource implements EntryInterface, \ArrayAccess
     /**
      * {@inheritdoc}
      */
-    public function offsetSet($name, $value)
+    public function offsetSet($name, $value): void
     {
         throw new \LogicException('Entry class does not support setting fields.');
     }
@@ -150,7 +150,7 @@ class Entry extends LocalizedResource implements EntryInterface, \ArrayAccess
     /**
      * {@inheritdoc}
      */
-    public function offsetUnset($name)
+    public function offsetUnset($name): void
     {
         throw new \LogicException('Entry class does not support unsetting fields.');
     }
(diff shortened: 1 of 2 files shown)


bullenb:
Currently the composer.json file specifies PHP 7.2 and above is supported. This would break the PHP 7.2 integration as `mixed` is not a valid return type until PHP 8.0.

Seppo-Konttinen-Solteq:
@bullenb hi. Can the 7.2 support be removed as I think it is not anymore supported and PHP 7.4 doesn't have active support and security support ends 28 Nov 2022. So it would make sense to do release without 7.x support and add the PHP8.1 compatibility fixes. Thanks.

bullenb:
@Seppo-Konttinen-Solteq That would be a decision for the Contentful team (@Sebb767 ?) I guess but I would be of the opinion that it would be good to remove the PHP 7.2 compatibility. The issue with that of course is that it's a breaking change that would require a major version release (assuming this package is using semver or some such versioning system...) along with checking compatibility of all the required dependencies.

Seppo-Konttinen-Solteq:
@bullenb Hi. Yes of course. Option would be to do major version release without PHP 7.x support while leaving the old one with it for some period of time. That would give people time to upgrade their system.

Sebb767:
We'll release a separate major version with only PHP 8.0+ support as soon as PHP 7.4 ends, so in roughly a month :)

This will also include support for cross-space references. We also discussed the option of supporting two major versions separately, but we decided against this for now, especially since the PHP 7 EOL date is not far off.

Seppo-Konttinen-Solteq:
Thanks for the info @Sebb767 and that is good news. However we are trying to test and release new version of Magento that depends to PHP8.1 by the time older versions with PHP7.4 are EOL. The older Magento versions happens to lose their support at the same time that PHP7.4 does. So is there any change that the new major version could be released sooner so we would have time for testing. Might be alot to ask but asking anyway 🙂

Sebb767:
I'll try to fast-track the release @Seppo-Konttinen-Solteq :) No promises, but we'll probably simply but a maintenance only-sign on the old release.

Seppo-Konttinen-Solteq:
@Sebb767 thanks! Approciate all the effort 🙂

Seppo-Konttinen-Solteq:
Hi @Sebb767. Any news of timetable as the PHP7.x support should have ended yesterday?

Seppo-Konttinen-Solteq:
@Sebb767  hi.
I still get this
Fatal error:  During inheritance of ArrayAccess: Uncaught Exception: Deprecated Functionality: Return type of Contentful\Delivery\Resource\Entry::offsetExists($name) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /app/vendor/contentful/contentful/src/Resource/Entry.php on line 137 in /app/vendor/magento/framework/App/ErrorHandler.php:61

should these be released already as PHP7 is not supported anymore?

Sebb767:
(Finally!) fixed with [v7.0.0](https://github.com/contentful/contentful.php/releases/tag/v7.0.0) :)

Seppo-Konttinen-Solteq:
Good news @Sebb767 . Thank you!

PR #318, opened 2022-10-30
Deprecation message for JsonSerializable

asuri0n:
> Method "JsonSerializable::jsonSerialize()" might add "mixed" as a native return type declaration in the future. 

Adding explicit `array` return type to suppress this message.

There is already [an open PR](https://github.com/contentful/contentful.php/pull/317) that change the return type to `mixed` but this type is only available in php 8 and the repo is php > 7.2.


Sebb767:
Fixed with [v7.0.0](https://github.com/contentful/contentful.php/releases/tag/v7.0.0) :)

